PR #3, opened 2016-01-03
!playing

Aidan0000:
Just a simple command !playing that only displays current playing song, helpful at times since sometimes the song title doesn't fit on the right hand side.


jhgg:
This should be implemented in develop branch. Will re visit when merging.

PR #28, opened 2016-01-20
Some command changes

iMxAndyy:
Changes to !shuffle so only the owner can perform the command. 
Added !next for owners to force skip the song
Removed that from !skip, to allow the owner to still vote to skip like everyone else.

### DIFF
--- a/bot.py
+++ b/bot.py
@@ -1,0 +1,732 @@
+import time
+import inspect
+import traceback
+import asyncio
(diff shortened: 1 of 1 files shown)


Neddz:
I think the permissions feature will be developed soon so I am not sure if this is necessary.

iMxAndyy:
fair enough

PR #29, opened 2016-01-20
Added support for SoundCloud sets (and potentially other playlists supported by youtube_dl) and changed the auto_playlist a little

Neddz:
The auto_playlist used to generate exception if it wasn't auto-summoned from the beginning. What's more, if summoned later, it doesn't work. I moved it to summon so we create the await there and it works in the previous scenarios as well as in the auto-summon one.

I added the ability to process SoundCloud sets and potentially other supported playlists that used to crash the server because the server tried to process the whole list as 1 song and when it started you were not able to add more songs, you had to reboot the server.

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -363,7 +363,7 @@ class MusicBot(discord.Client):
 
             reply_text = "Enqueued **%s** to be played. Position in queue: %s"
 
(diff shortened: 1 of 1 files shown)


imayhaveborkedit:
I'm going to add this change myself because other stuff broke merging for this.

PR #30, opened 2016-01-21
Whitelist URL

jonnyboy0719:
This will basically whitelist the URL's, so you can't break the bot (hopefully, since that's why I created this) with faulty url's, or un-trusted sites.

### DIFF
--- a/config/whitelist_music.txt
+++ b/config/whitelist_music.txt
@@ -1,0 +1,5 @@
+.youtube.com
+//youtube.com
+.bandcamp.com
+.soundcloud.com
+//soundcloud.com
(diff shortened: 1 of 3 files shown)


deansheather:
Also btw, you should be merging into `review`... Nice idea, though. :smile: :+1: Especially handy for blocking porn sites (`youtube_dl` supports some of them...) et. al.

jonnyboy0719:
ah crap, sorry. Was using the develop branch one. Can check the review branch later and do the same thing there.

And I have never heard of `fuzzywuzzy`, but I could look into it :) (Yeah, I'm still pretty new to Python)

deansheather:
:)

PR #40, opened 2016-01-25
Added LastFM Scrobbling. ONLY works if LastFM can resolve entry.title

dbkynd:
Using the entry.title string to query LastFM to get an artist and track to sent back to the scobbling API.
This avoids counting on a similar format since the track data might be <artist> - <track> or <track> ~ <artist>
If there is a more reliable way to get artist and track from say, the youtube_dl lib. That would be the way to go because LastFM doesn't resolve them reliably with a lot of special or non English letters. Though I've been surprised by some tracks that it has done correctly.

It sends a scrobble immediately on track play. Some other apps have scrobbling delayed for up to 10-30 seconds of continuous play for track skips. But I wasn't quite sure how to implement that.

### DIFF
--- a/musicbot/scrobble.py
+++ b/musicbot/scrobble.py
@@ -1,0 +1,78 @@
+import urllib
+import json
+import hashlib
+import time
+import sys
+
+sk = ''
+api_key = ''
+api_secret = ''
+
+class Scrobble(object):
+	def __init__(self, user, passwd, key, secret):
+		global sk
+		global api_key
+		global api_secret
+		api_key = key
+		api_secret = secret
+		api_sig = hashlib.md5(("api_key" + api_key + "methodauth.getMobileSessionpassword" + passwd + "username" + user + api_secret).encode('utf-8')).hexdigest()
+		params = {
+			'method': 'auth.getMobileSession',
+			'format': 'json',
+			'username': user,
+			'password': passwd,
+			'api_key': api_key,
+			'api_sig': api_sig
+		}
+		try:
+			sk = self.request(params)['session']['key']
+		except:
+			print('There was an error getting a session key with the LastFM credentials provided.')
+			sys.exit()
+			
+	def send(self, entry):
+		global sk
+		global api_key
+		global api_secret
+		params = {
+			'method': 'track.search',
+			'format': 'json',
+			'track': entry.title,
+			'api_key': api_key,
+			'limit': '1'
+		}
+		try:
+			t = self.request(params)['results']['trackmatches']['track']
+		except:
+			print("There was an error searching LastFM for '" + entry.title + "'")
+			return
+			
+		if len(t)==0:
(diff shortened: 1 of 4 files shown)


musicbot/scrobble.py
```diff
@@ -0,0 +1,78 @@
...
+
+sk = ''
+api_key = ''
+api_secret = ''
+
+class Scrobble(object):
+	def __init__(self, user, passwd, key, secret):
+		global sk
+		global api_key
+		global api_secret
```
Rapptz, line 15: Why are you using globals? Just do `self.sk = ...` and access it via `self`.

musicbot/scrobble.py
```diff
@@ -0,0 +1,78 @@
...
+			'username': user,
+			'password': passwd,
+			'api_key': api_key,
+			'api_sig': api_sig
+		}
+		try:
+			sk = self.request(params)['session']['key']
+		except:
+			print('There was an error getting a session key with the LastFM credentials provided.')
+			sys.exit()
```
Rapptz, line 31: You shouldn't exit the entire application because the login failed.

musicbot/scrobble.py
```diff
@@ -0,0 +1,78 @@
...
+import hashlib
+import time
+import sys
+
+sk = ''
+api_key = ''
+api_secret = ''
+
+class Scrobble(object):
+	def __init__(self, user, passwd, key, secret):
```
Rapptz, line 12: In Python you can't mix tabs and spaces and it's not recommended to do so inter-file mixing like this. Stick to spaces and don't use tabs.

musicbot/scrobble.py
```diff
@@ -0,0 +1,78 @@
...
+		except:
+			print('There was an error scobbling the track ' + entry.title + "'")
+			return
+		
+	def request(self, params):
+		url = "https://ws.audioscrobbler.com/2.0/"
+		params = urllib.parse.urlencode(params).encode()
+		r = urllib.request.Request(url, data=params, headers={"Content-Type": "application/x-www-form-urlencoded"})
+		u = urllib.request.urlopen(r)
+		return json.loads(u.read().decode('utf-8'))
```
Rapptz, line 78: This library is asynchronous. Using e.g. `urllib` is blocking and against the design of the bot. Use `aiohttp` instead and `async def`/`await`.

deansheather:
I thought of this idea as well, but this implementation can only scrobble to one Last.FM account. :disappointed:

imayhaveborkedit:
On top of these issues:
- PRs should be made to the review branch
- This is a feature for after V2 comes out

musicbot/scrobble.py
dbkynd, line 15: Literally my first experience with python. Still learning. I appreciate the quick response.

PR #47, opened 2016-01-29
Saves the songs the users queue in the backup list.

johnroot:
One requested feature on the server I'm using this bot on
(TKbreezy/ConeyZZ subscriber Discord) was that the bot play music that
was previously requested when there's nothing on the queue. This
accomplishes that by writing to the backup list when a user !plays a
song and erasing it from the backup list when that song is !skipped.

This seems to work, even though the bot complains that "the file is
locked" whenever a skip occurs.


imayhaveborkedit:
This should be an option or a command, or both.  I'll implement this at some point since its really quick to do, but PRs go to review, not develop.

PR #49, opened 2016-01-30
Change default behaviour to youtube-search.

Imvoo:
This change is so if arguments are passed to !play without a url, it will default to searching and playing the first video on YouTube that gets returned from those terms. Currently, if no url is passed to !play nothing is played.

Also, using `ytsearch:<songname>` doesn't work properly as the arguments get split off via spaces, i.e. `ytsearch:Lego House` just searches **Lego** instead of **Lego House** even if surrounded by quotes or apostrophes. The change to bot.py to combine multiple arguments solves this, but I'm not sure if it would cause other issues.

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -717,7 +717,13 @@ class MusicBot(discord.Client):
                     continue
 
                 if args:
-                    arg_value = args.pop(0)
+                    arg_value = None
+
+                    if len(args) > 1:
+                        arg_value = ' '.join(args)
+                    else:
+                        arg_value = args.pop(0)
+                        
                     handler_kwargs[key] = arg_value
                     params.pop(key)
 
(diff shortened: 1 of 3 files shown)


imayhaveborkedit:
This is pretty neat, but i'll have to see if there's a better way to do this than messing with argument parsing.  I might just make a command for it though, or rewrite this into a `!search` command with video selection using `wait_for_message`.

musicbot/bot.py
```diff
@@ -717,7 +717,13 @@ def _get_owner_voice_channel(self):
                     continue
 
                 if args:
-                    arg_value = args.pop(0)
+                    arg_value = None
+
+                    if len(args) > 1:
```
Neddz, line 722: Wouldn't this affect commands that have more than 1 argument? Which I believe would be the permissions ones or adding/removing from blacklist now.

imayhaveborkedit:
I have implemented this in review.  The current behavior is to combine the entire input into one argument and pass that to youtubedl, which should handle both using a url and searching for a video based on the input.

imayhaveborkedit:
I'm changing this to have its own search command now.  I found that there's 4 services you can search and you can request an arbitrary number of results.

PR #52, opened 2016-01-31
Added dynamic help message

bsquidwrd:
This will update on the fly as commands are added and removed


imayhaveborkedit:
I've been meaning to do something like this for a while.  I'll probably mess with this a bit to get it looking how I want it, but its a pretty decent start.

PR #61, opened 2016-02-02
Added comment line checking in utils.load_file()

deansheather:
This is especially handy for the `autoplaylist.txt` file, as server owners can add comments into their files and not worry about breaking the bot.

Using a nested `if not` statement to make sure that we don't run `line.startswith('#')` on an empty line. Not sure if this would be an issue or not, but it's safer? :joy:

### DIFF
--- a/musicbot/utils.py
+++ b/musicbot/utils.py
@@ -9,7 +9,8 @@ def load_file(filename):
             for line in f:
                 line = line.strip()
                 if line:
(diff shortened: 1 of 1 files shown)


imayhaveborkedit:
I'll add this manually since i'm going to force push soon.

deansheather:
Okay then.

PR #93, opened 2016-02-14
Added search feature to bot

kevinmost:
Bot uses the Youtube API to search for your message, and plays the first
search result. Add your API key in the Credentials section of the
options.txt, under the key "YoutubeAPIKey"

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -4,6 +4,9 @@ import traceback
 import asyncio
 import discord
 import sys
+import urllib
+from urllib import request
+import json
 
 from discord import utils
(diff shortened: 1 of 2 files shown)


kevinmost:
I saw talk of implementing a `!search` command on some other PR threads. I figured an "I'm feeling lucky" style search would be nice for a first release, and then adding a way to interactively decide on a song from the search results would be simple enough to add in the future.

imayhaveborkedit:
- Youtube-dl can already do this
- HTTP requests should be made with `aiohttp`
- Don't PR to develop
- This functionality has already been added in review.

PR #201, opened 2016-04-15
Add OSX documentation, move docs into own folder

henryzhangsta:
I added OSX installation documentation and pulled the docs off of the Wiki into the repository directly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,10 @@ MusicBot is a Discord music bot written in [Python](https://www.python.org "Pyth
 
 ### How do I set it up?
 
-- [Windows Tutorial](https://github.com/SexualRhinoceros/MusicBot/wiki/Installation-guide-for-Windows-7-and-up "Windows instructions")
-- [Linux Tutorial](https://github.com/SexualRhinoceros/MusicBot/wiki/Installation-guide-for-Ubuntu-14.04-and-other-versions "Linux instructions")
-- [Docker Tutorial](https://github.com/SexualRhinoceros/MusicBot/wiki/Installation-guide-for-Docker)
-- OSX Tutorial - coming soon, whenever someone decides to make one.
+- [Windows Tutorial](docs/windows.md)
+- [Linux Tutorial](docs/linux.md)
+- [Docker Tutorial](docs/docker.md)
+- [OSX Tutorial](docs/osx.md)
 
 ### Commands
 
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -1,0 +1,31 @@
+**THIS GUIDE IS FOR INSTALLING THE `review` BRANCH OF THE MUSIC BOT.**
+
+# Table of Contents
+
+- [Introduction](#introduction)
+- [Installation](#installation)
+  - [Step 1: Preperation](#step-1-preperation)
+  - [Step 2: Installation](#step-2-installation)
+  - [Step 3: Configuration](#step-3-configuration)
+
+# Introduction
+
+Docker is a way to virtualize the Musicbot. By using Docker you do not have to worry about any dependencies or whether it may run on your platform or not. Simply use the docker image and run the music bot.
+
+# Installation
+
+## Step 1: Preparation
+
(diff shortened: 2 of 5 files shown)


deansheather:
Why were the docs moved into a folder in the repo? This seems useless. The idea of having the docs on the wiki is so that everyone can edit them easily... I think the Mac OSX documentation should be added to the wiki, @henryzhangsta.

deansheather:
I asked Henry about this in Discord, and he said that it makes updating the docs easier to do alongside the code, and that the docs will be kept in forks.

I do agree these are good points, but I still believe that having a wiki is the way to go. Any opinions, @imayhaveborkedit and @abalabahaha?

henryzhangsta:
The idea that I'm working on is that stuff like installation instructions change less often than user instructions. If a pull request adds another dependency, thats a code change and the doc change could go into the repo with that, and preserves separation between branches. User documentation such as command usage, FAQs, known issues, etc, stay in the wiki because those can change even without a code change.

deansheather:
Eh, that's another good point. I'd still prefer for it to be open, though, and changing the links everywhere to the new format might take a while (probably gonna have to leave links in the wiki for it).

ImthePeach:
Henry does make a good point, but most of the users we get in the Rhino help server we usually tell to go on the latest version anyways. When they do this they are used to still going to the other urls posted in the #musicbothelp channel topic which guides to the Wiki.

deansheather:
I can easily fix the channel topics for things and post a message in #announcements if need be, the only factor affecting this really is _"do we want to move the docs around?"_.

ImthePeach:
Users rarely go and look at even the wiki is the issue. Would they really go and try to look through the repo-docs and try to do some digging or just give up and ask for help? Most have issues figuring out that the `OwnerID` is NOT the 4 digit discriminator. I don't know if this would be such a great idea...

Theoretically it makes sense, especially to the tech-savvy people, but in practice it could cause more of a headache for the help server.

kojitsari:
This is pointless, people don't read the documents on the wiki or the comments in files they _have_ to edit to make the bot work. The majority won't read help files regardless of if they're packed with the bot, the ones that would read them on the wiki. 

This is also annoying when you need to for the repo, and make a pull request when you make small changes.

jord-nijhuis:
The biggest reason to have to docs with the repository is that the docs are specific for the version of the bot. If you would try to run an older version of the bot, you cannot find the documentation anymore as only the most recent version is available. To still make sure users can find the documentation you could redirect them from the wiki to the latest version of the documentation in the repository.

TLDR: Move the documentation into /docs, redirect them from the wiki to a file in the /docs folder

@kojitsari Isn't that an advantage? In my eyes, every change needs to be approved before it is actually added to production. I was actually surprised when I could just edit the wiki as an outsider.

abalabahaha:
I'd say keep the docs on the wiki. We don't need to wade through a mass of 9001 PRs whenever a typo is discovered or something breaks.

Unless we want to support legacy releases, the argument about docs being specific to old versions of bots is invalid. It's more headache for the help server trying to assist people in multiple, retired versions of the bot.

Also, developers sometimes forget to update the docs when adding new stuff. Then, that's an extra commit, which means a clogged git history.

Also, "if a pull request adds another dependency," what's to say a pull request can't add another command? If that happens, why should install docs be kept in the repo when user docs suffer from similar problems?

jord-nijhuis:
Normally you would do a code review when you have a pull request, this should filter out all the typo's and the docs not being updated. Pull requests are not that big of a deal, and smaller ones can just be accepted with one click of a button. 

Keeping your docs in the repository is for the same reason you have your code in the repository, so that you can look back on it, and to make sure every version is available for whatever reason.

imayhaveborkedit:
Something something external docs but that shits hard.  I dunno what to do about this honestly.  Some of the wiki stuff needs rewriting anyways but I can't be arsed now.

deansheather:
Since MacOSX documentation has been added to the wiki, I am closing this PR.

PR #227, opened 2016-04-27
Added a custom status option

pamymaf:
Enabling NowPlayingStatus will have the bot change it's status to what
it is playing (overrides CustomStatus).
Enabling CustomStatus will have the bot change it's status to Status.
Status is what the CustomStatus would change it's status to.


jayktaylor:
Personally I feel like this isn't done in the best way, there's no point having two booleans and a string in the config for the custom status. 

Just have the custom status string in the options and if it's not empty then the status will be set permanently to that, otherwise it'll put the current played music in the status. 

I don't see why you wouldn't want the music in the status anyway, so having a boolean to turn that specific feature on or off is a bit silly - just let 'custom status' take precedence if it isn't empty.

pamymaf:
Looking at it, I have to agree with you on the second boolean.

As for why? For multi server bots that might not want the music of one server to be seen by others.

rrbailey89:
This needs more testing. It did not work out of the box for me. It does not seem to grab the data from options.ini for me. unless I copied the pertinent lines of code incorrectly.

this is my options.ini

```
; Have the status change when a new song starts playing.
NowPlayingStatus = no

; Have the status be set in the configuration to a custom message. NowPlayingStatus overrides this.
; Remember, all statuses start with 'Playing'
CustomStatus = The Air Guitar

```

In order for me to get a custom status I changed the value in config.py from 

```
    now_playing_status = True
    custom_status = 'with music.'

```

to 

```
    now_playing_status = False
    custom_status = 'The Air Guitar' 
```

in order to accomplish a custom status.

pamymaf:
I cloned the current review and only made the changes in my commits, and it works for me here and pulls from the `options.ini`. Do you have any other code modifications made, and can you try with just my code?

PR #275, opened 2016-05-17
Fixed Docker support

lukegg:
Myself and many others have had issues with the Docker install for MusicBot for awhile now, so I fixed it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,29 +1,36 @@
-FROM ubuntu:14.04
+FROM ubuntu:16.04
 
-MAINTAINER Sidesplitter, https://github.com/SexualRhinoceros/MusicBot
+# Original file created by @sidesplitter - however myself and others have had problems with their dockerfile recently so I fixed it with this.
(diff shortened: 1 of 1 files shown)


jord-nijhuis:
Thanks for fixing it! As soon as it gets merged I well push it to the Docker Hub (sadly I have no permissions on this repo, so I need to merge it back and then Docker Hub gets triggered). We might want to use the default Python image instead of building our own, but I am not sure if it matters much

jord-nijhuis:
Actually, now that I think of it, why did you replace the ADD with a git clone? Not only does this mean that building a custom-source one has become impossible unless you first push it to the repository, you also always get a build from the review branch. So if a want to build another branch, no use, since I will get a review branch anyway. Could you please change this back? @LukeClements 

You also added some packages like vim, this will increase the size of the image, but has no use for the musicBot, so I suggest that we remove that as well

lukegg:
I'm not that familiar with Docker - I just did what worked for me and what I presumed would work for others too. I can't remember why I added vim but yeah, you're right.

I'll close this and let you push something @Sidesplitter.

PR #278, opened 2016-05-19
Disconnect, Shutdown, and Restart Message Replies.

ADetectionAlgorithm:
Bot posts back to the server when the {}Disconnect, {}Shutdown, or {}Restart message is fired successfully.  This shows the operator that the bot did in fact recognize the command, something that isn't always apparent when not looking at console, especially if the bot needs to restart.

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -1793,9 +1793,7 @@ class MusicBot(discord.Client):
     async def cmd_restart(self, message):
         await self.safe_send_message(
             message.channel,
-            'Restarting...',
-            expire_in=30,
-            also_delete=message if self.config.delete_invoking else None
+            'Restarting...'
             )
         await self.disconnect_all_voice_clients()
         raise exceptions.RestartSignal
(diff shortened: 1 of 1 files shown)


musicbot/bot.py
```diff
@@ -1781,14 +1781,32 @@ def check(message):
 
 
     async def cmd_disconnect(self, server, message):
+        await self.safe_send_message(
+            message.channel,
+            'Disconnecting from %s...' % server,
+            expire_in=30,
+            also_delete=message if self.config.delete_invoking else None
```
imayhaveborkedit, line 1788: `expire_in` and `also_delete` are pointless since the bot will not maintain the state of the messages to delete them after restarting.

ADetectionAlgorithm, line 1788: Yeah, it works on the disconnect cmd, but not on the restart or shutdown cmds.  I originally left them because I didn't know if they would work without them, but after testing, they have been removed.

ADetectionAlgorithm:
Restart and Shutdown commands fixed to where they will output (message parameter missed originally).  Disconnect command fixed to output voice channel instead of server on disconnect.  Unnecessary parameters removed on restart and shutdown, as the bot will not maintain delete memory after restarting.  Thanks @imayhaveborkedit for calling that to my attention.

imayhaveborkedit:
Done in 5efcaf10e270e1c017ff1450bb69931f258afde4

PR #287, opened 2016-05-22
Make strings configurable

jayktaylor:
Currently up to date with: **1.9.5_1** (Last updated: `15th June 2016`)
### What does it do?

This feature has been requested a lot by users - the ability to configure strings and responses within the bot without knowing where they are in the code or destructively modifying the code.

This is my implementation of it. There may be better, cleaner methods, but this seems to work fine (besides a few minor bugs).

In my own and most other people's opinions, this is needed to allow individuality. Currently, people using this bot will more than likely have the same generic responses and output, with no way to make it "their own". This feature allows for that.
### How does it work?

Editing strings is done via `strings.ini`. If the strings aren't found/they are empty, they'll fallback to defaults. Strings can contain various **placeholders** which are indicated by comments above the option(s) and look something like `{song}`.

**As of https://github.com/SexualRhinoceros/MusicBot/pull/287/commits/06086fce251022e43d658e2c7b0f340fb756f627**, you can also configure the time until categories of messages ('Success', 'Error' and 'Help') get deleted from chat automatically.

**For the more technical:** This implementation uses `ConfigParser` (the same way as the `Config` and `Permissions` classes) to parse an `.ini` file.
### What strings can I modify?

You can modify most strings that are **bot responses**, such as `Enqueued **{songs}** songs to be played. Position in queue: {position}`. You can also modify _some_ **error responses**, although a few can not be customized for the sake of ease.

**You can't modify console output**. This is important for when there are issues with the bot. When you, for example, require help and come to the help server, usually your issue would be noted in the console. It's hard to anticipate what your issue might be if you've modified what the text is in the console window.

There may be room for improvement in disallowing other strings from being customized that can help with debugging, such as `Unable to change avatar: {exception}`.


LouieK22:
I love this idea. Looks great. If only the admins would accept it and fix the conflicts.... @SexualRhinoceros

lmmfranco:
Any chance we could get this PR approved?

PR #291, opened 2016-05-24
Force update_deps script to run with elevated perms

jayktaylor:
Resolves various common `Access is denied` errors that Windows sometimes throws

### DIFF
--- a/update_deps.bat
+++ b/update_deps.bat
@@ -1,5 +1,30 @@
 @ECHO off
 
+REM  --> Check for permissions
+    IF "%PROCESSOR_ARCHITECTURE%" EQU "amd64" (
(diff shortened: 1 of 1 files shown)


jayktaylor:
@imayhaveborkedit You said the other day it's down to opus not being able to update due to Python using it, however people still seem to need to run the script as admin when they don't have admin privileges on their PC or when for some reason 'Access is denied' to other files, despite Python not running. Myself and @ImthePeach have seen this issue.

It would probably be best if this was merged to force using elevated permissions regardless, to avoid stupid issues like the above.

ImthePeach:
This would help resolve a lot of the `I got this red text saying no permission when running the file what do I do?` type of questions since a lot of people seem to not have themselves as the admin on thier computer or have their settings all janky and making us have to tell them to run it as admin. I approve.

Issue #300, opened 2016-05-30
Is there a way to make the bot listen to me(owner) on private chat


omerbadawy11:

ADetectionAlgorithm:
See Pull Request #301 for a method of achieving this.

imayhaveborkedit:
See #301 for why this does not work.

PR #301, opened 2016-05-30
Allow PM with Owner

ADetectionAlgorithm:
I made this easy change as an example for Issue #300 for PM control of the bot.  It isn't 100% known what would happen to run music commands in PM, but this is mostly just here as an example for #300.


imayhaveborkedit:
This does not work.  Commands require a server context to function, which doesn't exist in DMs.

PR #306, opened 2016-06-01
New Commands: !undosong, !remove [x], !playnext

berkuhlee:
`!undosong` removes the last item from the playlist (simple deque.pop() )

`!remove [x]` removes the item in the playlist at the provided index.

`!playnext url` adds a song to the top of the playlist.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # RhinoBot: The music bot for Discord.
 
+EDIT: This is a fork from Music-Bot-Review 5/30/2016. New commands: !undosong, !remove [x], !playnext
+
+==============================
+
 MusicBot is a Discord music bot written in [Python](https://www.python.org "Python homepage"). It plays requested songs and if the queue becomes empty it will play through a list of existing songs.
 
 ### How do I set it up?
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -4,7 +4,6 @@ import time
 import shlex
 import shutil
 import inspect
-import aiohttp
 import discord
 import asyncio
 import traceback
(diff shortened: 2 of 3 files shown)


HeavenIsLost:
@berkuhlee :+1:

jayktaylor:
Firstly, pull requests should be directed to the `review` branch, not master.

Secondly, your commands are written in a really unclean way.

`!playnext` is pretty much the same code as `!play`, why not have a method of calling the existing `!play` function but making the necessary changes to put the song at the top of the queue?

`!undosong` is confusing to me. Not only should this command technically just be `!undo`, but changing  a boolean and using that to somehow measure if a song can be undone is odd.

`!remove [x]` is poorly done too. There's no try/except or a simple check to see if the index even can be removed. You're just asking for errors there.

@imayhaveborkedit is definitely going to work on these features and add them in at some point, but this PR could do with a lot of revision to make it at the standard we'd be wanting for the bot.

(P.S You're also changing the README file with unnecessary info, remove the added text :+1:)

berkuhlee:
yeah the `!playnext` is quite messy

`!undo` works too, I just wanted to differentiate it with other bots with `!undosong`. The boolean logic for `!undo` is definitely convoluted and very basic, but it works (although currently anyone can `!undo` another user's song)

edit: embarassing, I forgot to add `self.undo = True` inside of cmd_play's function when I updated to 1.9.5's code

Yeah I agree `!remove` is poorly done, although the bot seems to handle errors well

imayhaveborkedit:
The playnext and remove commands are pretty trivial to to do, and an undo command requires some sort of state or history tracking, depending on how I want to implement it.  Even that's not difficult to do, but i'll get around to these in due time.

PR #346, opened 2016-06-13
partial revert

dschwartz783:
self.http does not exist. Reverting to `self.headers['user-agent']`, which works.


imayhaveborkedit:
This was to go with a change in discord.py.  Update your lib.

PR #364, opened 2016-06-22
Bot announces who joins or leaves the channel in first text channel

xDazdingo:
Just did this because people on my server wanted it.
The bot announces people who join or leave the channel the bot is in inside the first text channel.


berkuhlee:
what do you mean first text channel?

imayhaveborkedit:
Uhhhh no.  You can have your bot do this but it doesn't really belong in the main bot.

PR #370, opened 2016-06-23
FI 1.9.5_1

bzsl:
I think this will bring my own project up to the current 1.9.5_1 version of the original project.


bzsl:
Nope, got it backwards. GitHub is confusing.

Issue #399, opened 2016-07-02
Docker wiki fix


jseed:
I am in the process of updating the wiki for the Docker installation/deployment and thought I would create an issue to discuss the current issues, since there seems to be no 'pull requests' for wiki changes.  

Currently, the guide instructs users to pull the container from Dockerhub, and not to build it themselves.  The current 'maintainer' set up this container as an automatic build from his own fork of the repository.  This meant that any time there was changes to this repository, he would have to resync his fork in order for the changes to be deployed to the container.   The issue is that at some point, he stopped syncing the fork, leaving the container outdated and broken.  

The container was created from a forked repository because you can only set up automated builds on Dockerhub for your own repositories, or repositories for an organization that you belong too.  ( You may also be able to set it up for repositories you have access too, not sure on this one)

To resolve this, I will update the wiki to instruct users on how to build the image themselves.  
If someone involved with the project wishes to set up an automated build for the main repo, I would be happy to provide instruction, or set it up myself (if given the permissions), just let me know.  

I will close this issue after a week or two, or once someone involved with the project acknowledges it.

ghost:
I just wanted to mention that I'm super new to docker, and these are the troubles I ran in to using the docker walkthrough:

The command `docker build -t MusicBot .` yields the following error: `invalid value "MusicBot" for flag -t: Error parsing reference: "MusicBot" is not a valid repository/tag`

Using `docker build .` works, however the image doesn't have a repository or tag this way, just an id. If you replace "MusicBot" in the final run & stop commands with the id obtained from building the image, everything works swimmingly.

jseed:
Interesting, I wonder why the tag isn't working, I will have to look into that.  Probably a silly little syntax error on my end.
I'm glad  to hear it works with the id though! The tag is really just for convenience and ease of use.

jseed:
You can't have capital letters in the image name, woops.
Updated the wiki to use the tag musicbot instead of MusicBot

jseed:
Closing this issue since it the new process has been confirmed working

PR #400, opened 2016-07-02
Move docker image to Alpine Linux

jseed:
Now that docker users must build the image themselves (see #399), it is a good time to look into optimizing the image.  Building off of Alpine Linux instead of Ubuntu yielded significant improvements in both build time, and image size

Build time:
Alpine - 2:30
Ubuntu - 7:30

Image Size:
Alpine - 280 MB
Ubuntu - 735.7 MB

This change would see no change in the build process on the users side, other than a different mount point for the config.  
If you would like to test this build, follow the same process as describe in the wiki, except use config:/usr/src/MusicBot/config as the argument to the -v option when running the image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,17 @@
-FROM ubuntu:14.04
+FROM alpine:3.4
 
-MAINTAINER Sidesplitter, https://github.com/SexualRhinoceros/MusicBot
+# Install Dependencies
+RUN apk update \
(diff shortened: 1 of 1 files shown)


JigSawFr:
+1

arthurbarros:
👍

imayhaveborkedit:
I'm going to commit this to review but I suppose I should figure out how this works for myself at some point.

jseed:
Sounds good, until then I can (hopefully) help you with any questions about the changes

PR #447, opened 2016-07-22
Config: Implemented DeleteNowPlaying option and fixed the DeleteMessages description.

RheaAyase:


musicbot/bot.py
```diff
@@ -388,7 +388,8 @@ def _delete_old_audiocache(self, path=AUDIO_CACHE_PATH):
 
                 async for lmsg in self.logs_from(channel, limit=1):
                     if lmsg != last_np_msg and last_np_msg:
-                        await self.safe_delete_message(last_np_msg)
+                        if self.config.delete_nowplaying:
```
jayktaylor, line 391: I'd suggest putting this check above line 387, otherwise the bot gets the logs from the channel first before even checking if it will do anything with them

RheaAyase, line 391: I've no idea what's going on in the rest of the code so I didn't want to if-it-out in case it's important...

387: `if last_np_msg and last_np_msg.channel == channel and self.config.delete_nowplaying:`  
I can quickly commit that if that's better.

jayktaylor, line 391: That seems like a better idea

RheaAyase:
[New pull request](https://github.com/SexualRhinoceros/MusicBot/pull/448)

PR #468, opened 2016-07-30
Two new commands, addrole and removerole

samuribe:
Two new commands, addrole and removerole, which allow users to assign and unassign server roles from themselves as long as they have permission to manipulate that role via permissions.ini.


jayktaylor:
This is a MusicBot, not a moderation or administrative bot. It doesn't need commands like this.

HeroCC:
Why did you remove the autoplaylist?

PR #492, opened 2016-08-07
Added universal shortcut and a music icon

lyjacky11:


jayktaylor:
This really isn't needed, if someone wants to make a shortcut they can do it themselves, and including an icon file for the shortcut is un-necessary too.

lyjacky11:
Ok thank you

PR #496, opened 2016-08-08
Regularly check for bot updates

gurchik:
# What does this PR change?
- Changes the two version strings in constants.py to a single variable
- Adds a "version" command to check for updates
- Adds a few more options and a new section to options.ini
- Adds feature to send a private message to the bot owner when updates are available
# What does this PR reference?

Resolves part of #1
# Additional notes

In order to prevent the bot from spamming the owner, they have the option to disable private messages entirely (in which case you'd only be able to check for updates by using the "version" command, and when the bot does send a notification to the bot owner about an update, it will prevent sending additional reminders until a certain time has passed.

### DIFF
--- a/config/example_options.ini
+++ b/config/example_options.ini
@@ -29,6 +29,7 @@ Token = bot_token
 ; I don't want any more "how do I get the OwnerID" questions.
 OwnerID = 000000000000000000
 
+
 [Updates]
 ; Whether the bot should check for updates automatically. If yes, the bot will 
 ; send a message to the Owner when an update is available. If no, updates will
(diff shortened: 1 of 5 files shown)


jayktaylor:
As it has been 2 months without activity on this PR, it will be closed. If you wish to PR this feature again, please merge your changes and PR to the `review` branch. 

The updating system you've tried to implement here will likely be added in the future. There doesn't need to be as many changes as this PR has though, just a simple notice in the console stating there's an update would suffice.

gurchik:
Why was there no activity on this PR for 2 months?

PR #533, opened 2016-08-27
Created a script to update the bot and dependencies for linux

AleanaA:
Nothing major

The first three lines update the bot, then pip if it can, and then the dependencies.

After updating these, the script clears console and prompts the user if they want to run the bot.

Feel free to tell me if it's bad or not, still new to creating bash scripts but I figured this might come in handy.

### DIFF
--- a/update_bot.sh
+++ b/update_bot.sh
@@ -1,0 +1,23 @@
+sudo git pull
+sudo python3.5 -m pip install --upgrade pip
+sudo python3.5 -m pip install --upgrade -r requirements.txt
+
+clear
+
+read -p "Update complete, would you like to run the bot? (yes/no)" reply
+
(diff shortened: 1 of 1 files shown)


ImthePeach:
Should make it so it checks if up to date first, then if they aren't updated run the updates. Otherwise I do like this because it could help keep the "Oh I was supposed to update_deps?" questions out of the help server. Or unless, check dependencies uptodate if up to date > automatically runs > if not up to date > updates > runs

jayktaylor:
This is a decent PR. A script similar to this will likely be added in the future, but this isn't something we are looking for currently. Also yes as Peach says, it makes sense to ensure the bot is  not up to date first. 

This would be better controlled through the actual Python script, rather than a separate bash file.

PR #543, opened 2016-08-28
Update bot.py

nl-tata:
Add support for aliases. Inputting an alias in !play will automatically replace the alias with the song_url defined by the alias. You can add aliases with !alias <aliasname> <songurl>. You can delete aliases with !delalias <aliasname>. You can save changes to aliases with !save.


ImthePeach:
Needs work anyways. _already talked to you about it in Rhino Bot Help server_

imayhaveborkedit:
This is a disaster and will be taken care of with playlists when I get to it.

PR #559, opened 2016-09-01
Fix commands link in README.md

dude22072:
It was using the page title "commands-list" instead of "commands".


dude22072:
Duplicate. Woops.

PR #604, opened 2016-09-14
Actually stays open if python isn't found.

UltraFireFX:
If python is not found, the dialog box shows up and disappears almost immediately, this allows for people to work out python was not found.


jayktaylor:
Please direct this PR to the review branch.

PR #612, opened 2016-09-19
Added cmd_playlist

Michael-Chandler:
New command formats malformed youtube playlist urls (ex. 1) if it matches or just passes it to self.cmd_play if it doesn't.

(ex.1) https://www.youtube.com/watch?v=PT2_F-1esPk&list=PLMC9KNkIncKtPzgY-5rmhvj7fax8fdxoj
(ex.1 fixed) https://www.youtube.com/playlist?list=PLMC9KNkIncKtPzgY-5rmhvj7fax8fdxoj

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -1,4 +1,5 @@
 import os
+import re
 import sys
 import time
 import shlex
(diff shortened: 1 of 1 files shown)


imayhaveborkedit:
Those aren't "malformed".  They're videos in a playlist.  If you want to add a playlist, use the actual playlist url, or even just the id works.

PR #636, opened 2016-10-01
MusicBot Output Alternative

bindingflare:
(2nd try using GitHub... here we go!)

I always looked at how the bot sent messages to both the cmd prompt and Discord and thought of making changes. Changes like:
- Using 101 Markdown (More than code blocks) and emotes.
- Making it other-spam-in-text-channel friendly (Messages stay for shorter when DeleteMessages is set to "yes" + some commands do not invoke message when DeleteInvoking is set to "yes").
- Make the bot look human!
- Debug mode (Looked more like the bot talking to him/herself!) and startup

This is the result, from using Notepad++ to using VS Code/ Studio for GitHub plugins)  :D

(Lots of unnecessary stuff, like that big text i made at the start... Sry about that >.<)


imayhaveborkedit:
If you want to do something like this you have to use the review version.  Master is not going to receive an update like this.

bindingflare:
What is the difference between the review branch and the master branch?

imayhaveborkedit:
The review branch is where ive been writing the code for the next update.

bindingflare:
kk... will work on it later.

ImthePeach:
also this... 
image

don't do this... This is unnecessary you don't need a change log inside the code....
also this...

image

"Nothing in my queuelist rn..."
RN? My seeing into the future powers are kicking in on the help server now: "hey gyes wuts does the `rn` mean in this? I don't git it can u's just esplain to me?"

Yeah don't do that please...

ImthePeach:
Not everyone knows what "rn" means or other silly jargon/acronyms for that matter. If you are going to add or change an out put it needs to be a proper grammatically phrased sentence.

bindingflare:
Noted

PR #653, opened 2016-10-09
Updated FFmpeg version

evieluvsrainbows:
I have updated the FFmpeg version to the latest Windows build that was released October 7th. Everything works A OK as I tested it before making this PR.


AraHaan:
the ffprobe.exe file can be safely removed as it is mostly not used on the standalone version of ffmpeg. That is unless the bot uses that file.

PR #665, opened 2016-10-12
Update .gitignore

TheTou:
prevent your autoplaylist.txt to get overwritten on update


TheTou:
All my carefully picked songs were gone after an update :'(

jayktaylor:
This is already in review. https://github.com/Just-Some-Bots/MusicBot/blob/review/.gitignore#L12

PR #670, opened 2016-10-16
URL BlackList

tutur1004:
Add cmd to add URL blacklist !

!BlackSound <url>

And the sound can't be added to queue


jayktaylor:
This should be an issue, not a pull request.

tutur1004:
Ho i see sorry ^^
(I'm FR for me all where just Tickets x)

PR #671, opened 2016-10-17
maxvolume permissions

KingGold171:
Added - MaxVolume in permissions (Limit how loud people can set the bots
volume, the number set in the permissions.ini is the limit)


KingGold171:
People on the main server i run the music bot on wanted the ability to check current volume.  But due to the volume being a single permissions node which gives access to check and set,   This has been split into two permissions nodes.   One  for checking, and one for setting.

jayktaylor:
Please can you direct changes to the `review` branch.

musicbot/bot.py
```diff
@@ -1464,15 +1464,26 @@ def check(m):
...
-            {command_prefix}volume (+/-)[volume]
+            {command_prefix}volume
+
+        Shows the current volume.
+        """
+
+        if not new_volume:
+            return Response('Current volume: `%s%%`' % int(player.volume * 100), reply=True)
+			
+    async def cmd_setvolume(self, message, player, permissions, new_volume=None):
```
jayktaylor: This shouldn't be it's own command. For efficiency, it should be a part of the `volume` command, like it has always been.

musicbot/bot.py
```diff
@@ -1482,7 +1493,12 @@ def check(m):
 
         except ValueError:
             raise exceptions.CommandError('{} is not a valid number'.format(new_volume), expire_in=20)
-
+			
+        if permissions.max_volume and new_volume > permissions.max_volume:
+            raise exceptions.PermissionsError(
+                "[ERROR] You are not able to set the volume above (%s)" % permissions.max_volume
```
jayktaylor: Probably remove the `[ERROR]` here, it already says `You do not have permission to use this command.` because of the exception that you're raising.

musicbot/permissions.py
```diff
@@ -98,7 +99,8 @@ def __init__(self, name, section_data):
         self.max_songs = section_data.get('MaxSongs', fallback=PermissionsDefaults.MaxSongs)
         self.max_song_length = section_data.get('MaxSongLength', fallback=PermissionsDefaults.MaxSongLength)
         self.max_playlist_length = section_data.get('MaxPlaylistLength', fallback=PermissionsDefaults.MaxPlaylistLength)
-
+        self.max_volume = section_data.get ('MaxVolume', fallback=PermissionsDefaults.MaxVolume)
```
jayktaylor: Remove the space after `section_data.get`

jayktaylor (changes requested):

musicbot/bot.py
KingGold171: What if i was to have the [Error]    is they have permissions to change it but try going above the set limit, and the default if they don't ?

musicbot/bot.py
KingGold171: Volume and Setvolume back as one command (MaxVolume permissions still included)

musicbot/permissions.py
KingGold171: Space removed

musicbot/bot.py
```diff
@@ -1482,7 +1481,12 @@ def check(m):
...
```
KingGold171: Bot.py changes based on Atom -- @jaydenkieran

KingGold171:
Alot of changes based on Atom 👍 Doesn't seem to break anything  @jaydenkieran

KingGold171:
Redone using the review branch as a base [CLOSED]

PR #905, opened 2017-02-05
Add announce command

PlanetTeamSpeakk:


Anidetrix:
Doesn't work. Messages look like this:
`<discord.message.Message object at 0x7f723c4b7828>`

Anidetrix:
Don't know about the `master` branch, but for `review` branch the following changes should be made:
1) `async def cmd_announce(self, *, message):` ->
    `async def cmd_announce(self, message, *, message_content):`
2) `await self.send_message(server.default_channel, message)` ->
    `await self.send_message(server.default_channel, message_content)`

jayktaylor:
We already document how to do this on the [modification page](https://github.com/Just-Some-Bots/MusicBot/wiki/Modification). The gist for it is [here](https://gist.github.com/jaydenkieran/c18ddef3817c58ad1cbb6018d33790eb). It won't be added as a native feature, at least not now anyway.

PR #962, opened 2017-02-25
Autoplaylist Improvement

ConnorPickles:


Humainenz:
Not sure if this is the right place, but allow the play of already downloaded files in your audio_cache folder. Maybe if a request is asked, it can check against the files already cache and play if in the audio_cache, otherwise download the audio?

PR #1059, opened 2017-04-19
Added pause/resume function that is called by typing the CommandPrefix

ianbeu:


jayktaylor:
There's no reason for us to add this into the bot, we already have `!pause` and `!resume`. We're not going to make it so that typing the command prefix on its own does this.

ianbeu:
My idea was to get a faster way to pause and resume music. Thats necessary when beeing ingame and you only want that the music plays between rounds.

jayktaylor:
We're totally cool with you modifying the bot in that way ([info](https://github.com/Just-Some-Bots/MusicBot/wiki/Modification)) for your own usage/your own fork. It just isn't something we'll add into the base version of the bot in this repo.

Thanks for the PR, though.

PR #1180, opened 2017-06-06
Add Dependency Autoupdate to Docker Installation

FatalMerlin:
This would prevent errors caused by Dependencies like youtube-dl.
The container should be run with --restart=unless-stopped !
Otherwise we need a capsulation script for the execution, because this would effectively kill the root process (PID 1 '/bin/sh -c python3.5 run.py').
Also fixes the MAINTAINER Tag.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,7 @@ WORKDIR /musicBot
 #Install PIP dependencies
 RUN sudo pip install -r requirements.txt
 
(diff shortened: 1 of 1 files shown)


jayktaylor:
A new Dockerfile was included in https://github.com/Just-Some-Bots/MusicBot/commit/ee8be2ac6af305a79b55cc51f4ddf29f2a6232bf.

PR #1412, opened 2017-12-04
Added cmd_playlist & YT playlist URL support

KingGold171:
cmd_playlist - load a pre-set .txt file to the song queue
allows for multiple autoplaylist.txt file with different names, different songs.

MAYBE soon allow to set which one will automatically plays if no one has requested anything.

Added support for YouTube playlist URLS in autoplaylist.txt - takes each song from the playlist and plays automatically as normal


KingGold171:
Well that failed and didn't compare the changes

PR #1442, opened 2017-12-28
README file translated to Korean

hth225:
I translated README.md file to Korean for someone who has language problems


MattBSG:
We’ve moved documentation, if/when we decide how/if we will do translations, we will look more into this (especially other languages, which aren’t translated at all at the moment).

jayktaylor:
Hey, we're probably not going to convert the README into different languages, however do feel free to create a PR for a Korean i18n file in this folder: https://github.com/Just-Some-Bots/MusicBot/tree/review/config/i18n.

There's some information about what this means on https://just-some-bots.github.io/MusicBot/#can-i-change-the-bots-responses.

PR #1449, opened 2017-12-31
Review - Time and Date Based logging system

OzzyHelix:

- [X ] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [ X] I have tested my changes on Python 3.5/3.6

----

### Description
The logging system that writes the old log to a file called musicbot.log.last 
I found a better way of doing it it still has musicbot in the file name but now keeps logs based on time and date picture example 
b5b79b70f4b2b9a3ac049553afe613a2

no issues as far as I can tell


OzzyHelix:
please ignore 
8db8b37  not sure why that's there

OzzyHelix:
@jaydenkieran  I'm very sorry I forgot to fix the logging message in the sanity checks its fixed now
image

OzzyHelix:
the if statement was causing the log message not to pop up so sorry I have fixed that now

jayktaylor:
The issue with this is that it has no cleanup currently. People could end up with tons of log files because the old ones aren't deleted, in contrast to what happens currently. I'm not entirely convinced that the current system isn't suitable enough to keep around. Maybe if you addressed these concerns and write some code that deletes logs that are older than perhaps a day, it would be a more suitable change.

OzzyHelix:
I suppose thats fair

OzzyHelix:
I should be able to address the issue

OzzyHelix:
Ok I have fixed it the reason I haven't released a merge on my fork is because I've only tested it on Windows and so I've setup a Linux VM to make sure it works on that if it does then I'll assume it will work on OSX too

OzzyHelix:
I screwed up the fix works for python3.6 but not python3.5

OzzyHelix:
ok nvm I figured out the problem gonna test it on ubuntu and see what happens than I'll push it up to github

OzzyHelix:
as promised it will now delete logs after one day and this works on python3.5 and python3.6

OzzyHelix:
sadly I spent the whole night on this so I'm a bit tired but there you go I also agree that my logging system before the one day delete did make too many logs for weeks I'm been trying to fix that myself kind of glad I did

OzzyHelix:
I will be honest I'm glad you asked for that change. I've been meaning to fix it so it didn't make a million logs I just haven't had the drive to fix that as of late

OzzyHelix:
the concerns were addressed in 
3a7b62a it does delete logs if their older than a day however it counts it in seconds so a day is about 86400 seconds

musicbot/bot.py
```diff
@@ -226,6 +223,21 @@ def _setup_logging(self):
             dhandler.setFormatter(logging.Formatter('{asctime}:{levelname}:{name}: {message}', style='{'))
             dlogger.addHandler(dhandler)
 
+# Here's Where log clean up comes in
+    now = time.time()
+    cutoff = now - (1 * 86400)
```
MattBSG, line 231: `time.time()` should be an `float` type. Is there a particular reason for multiplying `1` and `86400`?

OzzyHelix: I was working within my current skills set with python multiplying by 1 and 86400 will always result in the number coming out as 86400 which in seconds is a day entering the day in seconds did not work for me it had to be a math problem so to make old logs delete after a day I just had it multiply 1 * 86400 so it would always delete the old logs after a day

OzzyHelix: by simply putting 86400 it would not work correctly so I made it a multiplication problem that would result in the logs being deleted after a day

OzzyHelix: now = time.time() all this code to see what time it is if it sees the a file's age exceeds 86400 seconds it would delete those log files so it cleans up everything this code is run the way I did it however means the bot needs to be restarted in order to delete the old logs

OzzyHelix: I've also noticed a side effect of running this on Windows it deletes logs based on their creation date so depending on how many times someone restarts their bot in a day it will result in them either having a lot of logs or little it depends on how many times the bot is restarted a day

OzzyHelix: also I've tested the code very thoroughly it does work on python 3.5 and 3,6 and I agree with  jaydenkieran the system you guys have right now isn't suitable enough to keep around so this code does create logs based on time and date and in order to start the clean up someone has to restart their bot

OzzyHelix:
forgive me but I couldn't find a better place to propose an idea  perhaps it would be cool to use the spotify integration recently introduced into discord to act as a now playing status again sorry could not find a better place to give that idea

OzzyHelix:
I was going to ask if there are any changes to this that may be required I can tell you this system does work I've tested it for about a month now and it will delete messages older than a day but to cause that people would need to restart their version of the bot

OzzyHelix:
I screwed up let me fix this

OzzyHelix:
nvm I thought I screwed up

OzzyHelix:
Hey if its ok with you guys I'm going to close this pull request and reopen it along with another one that is unrelated to this one

Issue #1482, opened 2018-01-23
Bot keeps playing in empty channel [Raspberry pi]


ThatzOriginal:
Please make sure these boxes are checked (put an 'x' inside them) **before** creating the issue.

- [X] I am using Python 3.5 or higher (run `python --version` on the command line)
- [X] I have followed the [official guides](https://github.com/Just-Some-Bots/MusicBot/wiki) to install the bot for my system
- [X] I have updated my dependencies to the latest version using the appropriate `update_dependencies` script
- [X] I am running version 1.9.6_2 of MusicBot

----

### Description of issue

Bot doesn't auto pause when in an empty channel. (I also reinstalled rasbian and the bot onto the Pi just to make sure and issue still persisted)

### Steps to reproduce

1. Install the bot on a raspberry pi

2. Set in options pause when empty channel.

3. Join the channel with the bot in it.

4. Leave the channel (bot keeps playing with empty channel.

----

### Log file
Please attach your MusicBot log file (located at `logs/musicbot.log`) to this issue. You can do so by dragging and dropping the file here.

[musicbot.log](https://github.com/Just-Some-Bots/MusicBot/files/1658014/musicbot.log)

itsTheFae:
This is due to the checking for empty channels (`VoiceStateUpdate`) being a little funky, as well as `on_player_finished_playing()` not setting up a check for empty channels when it queues a new song (like the check used in `_join_startup_channels()`)  

I've patched part of this issue but haven't added the checking to `on_player_finished_playing()` as I'm almost sure there is a better method than just copying the lambda used in `_join_startup_channels()` but I am not the person to make that call.  Keeps the patch work a tad more simple too.. 

Devs can find the related code in a cloned repo here:  https://github.com/itsTheFae/DiscordMusicBot/tree/patch-auto-pause

MattBSG:
@itsTheFae Thanks for the work, if we’d like anything updated we can request changes in a PR. No harm in submitting one though :)

jayktaylor:
#1484 should have fixed this.

Issue #1477, opened 2018-01-22
Autoplaylist is not checking user present on start


falcon4fun:
I have found quite an old bug. It's like a bot doesn't verifies if there is somebody in the voice while using:
1. Autoconnect to channel on start
2. Autoplay enabled
3. Autopause enabled

I have found that every time when bot starts/relaunches, it starts to play music on all servers even nobody in the voice. And the solution is enter to voice and exit, then bot see that
Can you try reproduce? (: If ya, add initial checking. I tryed to do it by my self, but failed (: 

To reproduce:
1. Enable ^ 3 options
2. Add some tracks to autoplaylist.txt
3. Launch bot
4. Go to your TEXT channel and write !np and see, that music is playing but it should not :)
Tryed on review build. Still present.. :)
[https://i.imgur.com/bV88mLx.png](https://i.imgur.com/bV88mLx.png)

https://cdn.discordapp.com/attachments/252967574168535040/404804293648973835/unknown.png
https://pastebin.com/hMrNFYYT

itsTheFae:
I've noticed this as well.    
There is a check coded in `_join_startup_channels()` but the conditions as they are now have a little issue which prevents the one-shot lambda from being run.  I've created a patch for this as well as Issue #1482 targeting the latest review branch.   

Devs can find the patch commits here:  https://github.com/itsTheFae/DiscordMusicBot/tree/patch-auto-pause

falcon4fun:
I have found that sometimes it works.. I disabled equalisation.

In log:

> Initial autopause in empty channel
> Initial autopause in empty channel

itsTheFae:
In my case, the bot is always restoring a serialized queue when it starts.  I suspect the restoration of the queue might contribute to the auto-pause failure condition on this issue as well as mine..   
Not totally sure, I trust the devs here would know better than me.

jayktaylor:
Should've been fixed by the above PR.

PR #1484, opened 2018-01-26
Patch auto pause in various cases.

itsTheFae:

----

### Description
The Auto-Pause feature has a few conditions that prevent automatically pausing the playing track when a channel becomes empty or has been empty when the bot starts.  This patch attempts to address several of those cases. 

Auto Pause on Click-&-Drag now works.
Auto Pause on member moving channels now works.
Auto Pause on Initial join now more reliable.
Auto Pause on playing a new track via autoplaylist now works.
 
This patch also resumes any paused players to prevent hanging ffmpeg threads on shutdown and restart calls.

#1482 and #1477

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -655,6 +655,14 @@ class MusicBot(discord.Client):
         await self.update_now_playing_status()
 
     async def on_player_finished_playing(self, player, **_):
+        def _autopause(player):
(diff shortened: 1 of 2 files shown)


MattBSG (approved):
Thanks for this PR, I've verified everything works as it should with your patch.

PR #1493, opened 2018-01-31
A maybe working restart command

AutumnClove:

- [x ] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [x ] I have tested my changes on Python 3.5/3.6

----

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -2546,10 +2546,15 @@ class MusicBot(discord.Client):
         return Response("Disconnected from `{0.name}`".format(server), delete_after=20)
 
     async def cmd_restart(self, channel):
-        await self.safe_send_message(channel, "\N{WAVING HAND SIGN}")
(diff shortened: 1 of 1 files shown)


ghost:
Maybe you could implement a check if it's Windows, Mac or Linux, because .bat files are Windows-only.

MattBSG (changes requested):
Support for other OSes

jayktaylor (changes requested):
This isn't going to work. Raising an exception will terminate the function, thus a subprocess will never start. A subprocess is not really going to be suitable, because CREATE_NEW_CONSOLE will just open a new command window. And, as pointed out by others, this will only "work" on Windows.

AutumnClove:
haaalp I can't get it to go back to tab size 4 ;-;

MattBSG:
Author moved issue to #1494

PR #1494, opened 2018-01-31
Fixed restart command

AutumnClove:

----

### Description
I already did a pull request for this but I messed up the indent sizes and couldn't figure out how to fix it. 
So I just did a new one .-.

#1493

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -2556,7 +2556,7 @@ class MusicBot(discord.Client):
         await self.disconnect_all_voice_clients()
         from sys import platform as _platform
         if _platform.startswith("linux"):
-			p = subprocess.Popen(runbot.linux.sh, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+	    p = subprocess.Popen(runbot.linux.sh, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
         elif _platform == "darwin":
             p = subprocess.Popen('runbot_osx.command', creationflags=subprocess.CREATE_NEW_CONSOLE)
         else:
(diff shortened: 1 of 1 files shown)


musicbot/bot.py
```diff
@@ -2546,9 +2546,22 @@ def check(message):
...
+        """
+        Usage: 
+            {command_prefix}restart
+
+        Restarts the bot, making it leave all voice channels, this also opens a new command prompt, with any saved changes to files.
+        """
         await self.safe_send_message(channel, "\N{WAVING HAND SIGN}")
         await self.disconnect_all_voice_clients()
-        raise exceptions.RestartSignal()
+        from sys import platform as _platform
```
jayktaylor, line 2557: bot.py already imports the sys module, you don't need to do this.

musicbot/bot.py
```diff
@@ -2546,9 +2546,22 @@ def check(message):
...
+        if _platform.startswith("linux"):
+	    p = subprocess.Popen(runbot.linux.sh, shell=True, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        elif _platform == "darwin":
+            p = subprocess.Popen('runbot_osx.command', creationflags=subprocess.CREATE_NEW_CONSOLE)
+        else:
+            p = subprocess.Popen('runbot.bat', creationflags=subprocess.CREATE_NEW_CONSOLE)
```
jayktaylor, line 2563: This isn't how subprocesses work. You're just going to end up creating multiple processes within the same process, and then you'll eventually run into memory errors.

musicbot/bot.py
```diff
@@ -2546,9 +2546,22 @@ def check(message):
...
+        raise exceptions.TerminateSignal()
```
jayktaylor, line 2564: I don't think this point will ever actually be reached until the subprocesses have finished, which again is a problem that contributes to the memory issues mentioned beforehand.

jayktaylor (changes requested):

MattBSG (changes requested):
Subprocesses and threads are not a good solution because it can create huge problems especially pertaining to memory. It’s also really bad practice to make tens of sub-threads in cases a user may never actually close the bot. A completely different solution will be needed (think how ours is now, but add on to it)

Issue #129, opened 2016-03-08
When using "!help", only list the commands the user currently has permission for


Ivoz:
Hopefully the title is self explanatory.

Or, alternatively, give the commands the user has permission for in a first line, and the ones they don't in a second line.

imayhaveborkedit:
Yeah yeah I know, the current help command is just sort of a place holder until I make a proper one.  The old was was literally just a link to the commands page.

MattBSG:
This has been introduced in #1495 and was merged to review. It will be included in the next release. Gonna go ahead and close.

PR #1495, opened 2018-02-01
Help command only returns commands you can use

MattBSG:

----

### Description
Changes the behavior for the help command. `{}help` now only returns commands you can run with your permissions. You can bypass this by running `{}help all` or if you are owner.

Feature request #129

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -1214,7 +1214,7 @@ class MusicBot(discord.Client):
         player.autoplaylist = list(set(self.autoplaylist))
         return Response(self.str.get('cmd-resetplaylist-response', '\N{OK HAND SIGN}'), delete_after=15)
 
-    async def cmd_help(self, channel, command=None):
+    async def cmd_help(self, message, channel, command=None):
         """
         Usage:
             {command_prefix}help [command]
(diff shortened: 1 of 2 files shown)


musicbot/bot.py
```diff
@@ -1223,30 +1223,49 @@ def _gen_embed(self):
...
-                        dedent(cmd.__doc__)
-                    ).format(command_prefix=self.config.command_prefix),
-                    delete_after=60
-                )
-            else:
-                raise exceptions.CommandError(self.str.get('cmd-help-invalid', "No such command"), expire_in=10)
-
-        else:
-            commands = []
+        async def gen_cmd_list(message, all=False):
```
jayktaylor: Not sure that this should be within the scope of the cmd_help function. `all` should probably be renamed too as it is the name of an built-in function.

musicbot/bot.py
```diff
@@ -1223,30 +1223,49 @@ def _gen_embed(self):
...
 
             for att in dir(self):
                 if att.startswith('cmd_') and att != 'cmd_help' and not hasattr(getattr(self, att), 'dev_cmd'):
```
jayktaylor: There's probably no reason to actually exclude the help command here, even though they're using it to check these commands. Including the help command ensures granularity.

musicbot/bot.py
```diff
@@ -1223,30 +1223,49 @@ def _gen_embed(self):
...
+                    user_permissions = self.permissions.for_user(message.author)
                     command_name = att.replace('cmd_', '').lower()
-                    commands.append("{}{}".format(self.config.command_prefix, command_name))
+                    if command_name in user_permissions.command_whitelist and not command_name in user_permissions.command_blacklist or all:
+                        self.commands.append("{}{}".format(self.config.command_prefix, command_name))
 
-        desc = '```\n' + ', '.join(commands) + '\n```\n' + self.str.get(
+        if command:
+            if command == 'all':
```
jayktaylor: Maybe convert the command to lowercase incase someone types `!help ALL`?

musicbot/bot.py
```diff
@@ -1223,30 +1223,49 @@ def _gen_embed(self):
...
+                        delete_after=60
+                    )
+                else:
+                    raise exceptions.CommandError(self.str.get('cmd-help-invalid', "No such command"), expire_in=10)
+
+        elif message.author.id == self.config.owner_id:
+            await gen_cmd_list(message, all=True)
+
+        else:
+            if self.commands == []:
```
jayktaylor: `if not self.commands` is probably a better way to validate this.

musicbot/bot.py
```diff
@@ -1223,30 +1223,49 @@ def _gen_embed(self):
...
+                raise exceptions.CommandError(self.str.get('cmd-help-no-perms', 'You don\'t have permission to use any commands. Run `{}help all` list every command anyway'), expire_in=10)
```
jayktaylor: "Run `{}help all` list every command anyway"?

musicbot/bot.py
```diff
@@ -1223,30 +1223,49 @@ def _gen_embed(self):
...
+            await gen_cmd_list(message)
+            desc = '```\n' + ', '.join(self.commands) + '\n```\n' + self.str.get(
+                'cmd-help-response', 'For information about a particular command, run `{}help [command]`\n'
+                                     'For further help, see https://just-some-bots.github.io/MusicBot/').format(prefix) + self.str.get(
+                'cmd-help-all', '\nOnly showing commands you can use, for a list of all commands, run `{}help all`')
```
jayktaylor: This line could be added conditionally on line 1268/1269 rather than having the same code (setting the desc variable and then returning) twice.

jayktaylor (changes requested):
Seems good, just a few minor things that could be changed.

jayktaylor (approved):

PR #1506, opened 2018-02-13
Config

xnaveira:

- [x ] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [x ] I have tested my changes on Python 3.5/3.6

----

### Description

First of all, this is not a pull request :) I just wanted to show you the idea in case someone wants to pick it up. Since I run this in a docker that can get rescheduled to a different server at anytime (i'm using docker cloud)  i don't want it to have state so instead for having a volume with config files i created an small script that sets the values in the files using environment variables. Now, the right thing to do would be to have the configuration options to be parsed by the argparse module and override the ones parsed by configparse but I really couldn't bring myself to do it.

Great job with the bot btw!!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.pyc
 ~*/
 .vscode/
+*.env
 
 audio_cache/
 dectalk/
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,16 +2,14 @@ FROM alpine:3.4
 
 # Install Dependencies
 RUN apk update \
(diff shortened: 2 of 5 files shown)


MattBSG:
We usually ask that all PRs are sent to review, as we will not merge to master. We also have full docker support up and running in review so if you believe you want to actually suggest these changes PR to review. Thanks :)

xnaveira:
Oh shit, I didn't get that you wanted the prs against review too. Sorry about that.

PR #1515, opened 2018-02-25
Beta

kosugikun:

----


MattBSG:
All PRs **must** go to review; an example of why is because you *can* make a translation for for i18n in Japanese. Never edit the strings directly in bot.py as they are **fallback strings** if an invalid i18n is specified or en.json does not exist. Also please do not translate our .md documentation files, if you do that it makes them what you set them to for all languages. If you have further questions please ask. Also please do not ever update the repository link in any file away from ours.

Issue #1488, opened 2018-01-29
[Feature request] Command Aliases


Leomelonseeds:
Sorry for submitting another request so soon, but I suddenly had an idea...
Create command aliases, and/or an aliases.yml/aliases config file option to manage them.
Examples: 
!p = !play
!bl = !blacklist
!stop = !pause
!q = !queue
!next = !skip
etc...
I think this is a good idea to improve the usability of this bot :D

Zenrac:
It should be easy with discord.Bot, but the bot is using discord.Client. You have to create a new fonct with the alias name that return the real command :
```
async def real_command(self, player, message):
    Some codes 
```
```
async def alias_command(self, player, message):
    return await self.real_command(player, message)
```
An other way can be to edit the message content before calling the command, but I'm not sure if it's a good idea.

MattBSG:
They specifically asked to be able to change it with config files, not editing the bot which is the whole point of the issue. Thank you for trying to help however

owocado:
I would also like to request the same feature. I thought it would be better to reply to this than to create a new ticket. :P

SternRowboat:
Zenrac's comment is helpful for people with a small amount of coding knowledge, like me.
Being able to do shorthand commands like p for play is really nice.
A user-friendly way of enabling this would be great!

haydster7:
100% agree this would be awesome

Since I've been using the bot I always add 4 or 5 blocks at the top of the bot python file that pretty much does this, but a config file would be so good, would make it easier to update without possible merge issues and config could be easily changed for people without coding experience

obi23lipnik:
I think one way to go about this is have a class made for all the commands, then a function reading from the config file could just `setattr` the new alias functions in.. Shouldn't be too hard.

EDIT: Just opened a pull request with this feature... Check it out => #1547

PR #1547, opened 2018-03-21
WIP: Aliases Feature

obi23lipnik:

- [X] I have tested my changes on Python 3.5/3.6

----

### Description
Adds aliases functionality with addition of `aliases.json` to config file, which is used to build alias function names before initialization.

#1488

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ config/autoplaylist.txt
 config/autoplaylist_removed.txt
 config/whitelist.txt
 config/blacklist.txt
+config/command_aliases.json
--- a/config/example_command_aliases.json
+++ b/config/example_command_aliases.json
@@ -1,0 +1,4 @@
+{
+    "play": ["p", "poo"],
+    "blacklist": ["bl"]
+}
--- a/musicbot/config.py
+++ b/musicbot/config.py
@@ -306,12 +306,10 @@ class Config:
             log.debug('The file \'command_aliases.json\' does not exist.')
         else:
             try:
-                command_aliases = json.load(open(self.command_aliases_file))
+                return json.load(open(self.command_aliases_file))
             except json.decoder.JSONDecodeError:
                 raise HelpfulError('Your command_aliases.json file is not a valid JSON!',
                                    'Properly configure config/command_aliases.json and re-run the bot.')
-
-            return command_aliases
 
     def setup_command_aliases(self):
         command_aliases = self.command_aliases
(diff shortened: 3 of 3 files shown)


musicbot/bot.py
```diff
@@ -75,11 +76,13 @@ def __init__(self, config_file=None, perms_file=None):
 
         self.blacklist = set(load_file(self.config.blacklist_file))
         self.autoplaylist = load_file(self.config.auto_playlist_file)
+        self.aliases = json.load(open(self.config.command_aliases_file))
```
jayktaylor: This isn't really good enough. This will stop execution if it throws an exception, such as a `JSONDecodeError` (for example, when the file isn't valid JSON). There needs to be additional checks for this. It may also be better to handle it in `config.py` instead and sanity check it there.

config/command_aliases.json
```diff
@@ -0,0 +1,3 @@
+{
```
jayktaylor: This file should probably follow the lead of the other config files and be renamed to something like `command_aliases_example.json` , with `command_aliases.json` being ignored in `.gitignore`. That way, we don't overwrite the command aliases configuration when there is a new bot update.

config/command_aliases.json
```diff
@@ -0,0 +1,3 @@
+{
+    "blacklist": "bl"
```
jayktaylor: It's probably better to interpret this differently to avoid redundancy:

```json
{
    "blacklist": [
        "bl",
        "whatever"
    ]
}
```

musicbot/bot.py
```diff
@@ -183,6 +186,19 @@ def _delete_old_audiocache(self, path=AUDIO_CACHE_PATH):
 
         return True
 
+    def _setup_aliases(self):
+        if len(self.aliases) > 0:
+            for command_name, alias in self.aliases.items():
+                if hasattr(self, 'cmd_{}'.format(alias)):
+                    log.debug('Command named {} already exists!'.format(alias))
```
jayktaylor: Let's be more descriptive with this, and change it to a warning instead - people should be aware if there is an issue with their aliases, rather than it being hidden away in the log file.

jayktaylor (changes requested):

musicbot/bot.py
obi23lipnik: Moved aliases setup to config.py and added checks there

config/command_aliases.json
obi23lipnik: Created `example_command_aliases.json` and added `config/command_aliases.json` to `.gitignore`

config/command_aliases.json
obi23lipnik: Changed the format as suggested

musicbot/bot.py
obi23lipnik: Now raises `HelpfulError`s to help the end user

leoTiGRiS:
is it still working tho?

jayktaylor:
superseded by #1809

Issue #841, opened 2017-01-16
RuntimeError: PyNaCl library needed in order to use voice


FireFrostLike:
Attempting to autosummon...
Found owner in "AAN", attempting to join...
Traceback (most recent call last):
  File "E:\Programs\Python\D\lib\site-packages\discord\client.py", line 307, in _run_event
    yield from getattr(self, event)(*args, **kwargs)
  File "E:\Programs\Python\sed\MusicBot-master\musicbot\bot.py", line 710, in on_ready
    owner_vc = await self._auto_summon()
  File "E:\Programs\Python\sed\MusicBot-master\musicbot\bot.py", line 149, in _auto_summon
    await self.cmd_summon(owner.voice_channel, owner, None)
  File "E:\Programs\Python\sed\MusicBot-master\musicbot\bot.py", line 1325, in cmd_summon
    player = await self.get_player(author.voice_channel, create=True)
  File "E:\Programs\Python\sed\MusicBot-master\musicbot\bot.py", line 362, in get_player
    voice_client = await self.get_voice_client(channel)
  File "E:\Programs\Python\sed\MusicBot-master\musicbot\bot.py", line 249, in get_voice_client
    voice_client = VoiceClient(**kwargs)
  File "E:\Programs\Python\D\lib\site-packages\discord\voice_client.py", line 213, in __init__
    raise RuntimeError("PyNaCl library needed in order to use voice")
RuntimeError: PyNaCl library needed in order to use voice

ghost:
not an issue, install PyNaCl

LazyHatGuy:
Tried that
 pip install pynacl
Requirement already satisfied: pynacl in /usr/local/lib/python3.5/dist-packages
Requirement already satisfied: six in /usr/local/lib/python3.5/dist-packages (from pynacl)
Requirement already satisfied: cffi>=1.1.0 in /usr/local/lib/python3.5/dist-packages (from pynacl)
Requirement already satisfied: pycparser in /usr/local/lib/python3.5/dist-packages (from cffi>=1.1.0->pynacl)
Still getting the error

ghost:
well, it probably isn't installing in the right place, or there is a different traceback, ask in the API Discord

hansmok:
Did you try "pip list" and see if it's with the other modules?

PR #1578, opened 2018-04-15
Docker: PyNaCl needs libsodium-dev

cocoastorm:

----

### Description
Turns out in the Dockerfile `libsodium-dev` is listed as a build dependency. So it's removed after the Dockerfile is finished building the Docker image. Python needs libsodium-dev to be installed in the image to [import nacl.secret properly (from Discord.py)](https://github.com/Rapptz/discord.py/blob/async/discord/voice_client.py#L59)

I wrongly assumed PyNaCl wasn't being installed with pip, but turns out manually importing `nacl.secret` leads to this error in the Docker image:

`Error loading shared library libsodium.so.23`

Adding `libsodium-dev` as a core dependency in the Dockerfile fixes the problem.

#841


Hydrar:
I discovered this too yesterday, although you only need to add libsodium to the dependencies apk install, -dev is only needed for the build portion.
Also what is up with all these tons of commits for this simple pull request? Just found this since I was about to put one up myself

cocoastorm:
I don't know actually! The pull request template mentioned how the review branch had to be based on it? Which was why I rebased my branch to be based on the review branch instead of the master branch.

If you want you can make your own PR and I'll close this one haha, since you mentioned how including libsodium itself without -dev part might be a better idea lol.

jayktaylor:
The review branch is the branch we develop on which is why we ask that PRs target that branch.

cocoastorm:
@jaydenkieran, hey is this PR okay? I'm a bit puzzled why all these commits came through. (Probably, because of my rebase) If I need to redo my PR I'm fine with that.

Thanks!

vdeville:
Thanks !

PR #1596, opened 2018-04-25
Change run.sh to use python3.5 instead of python3 

AleanaA:

- [ x ] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [ x ] I have tested my changes on Python 3.5/3.6

----

### Description
Keeps things uniform between run.sh and update.sh, as they should be the same, otherwise issues could arise such as dependencies being installed to a version of python associated with `python3` that isn't associated with `python3.5`

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-(command -v python3 >/dev/null 2>&1 &&
+(command -v python3.5 >/dev/null 2>&1 &&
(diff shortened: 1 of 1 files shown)


jayktaylor:
MusicBot supports Python 3.5 and higher. This pull request would be restricting that. If anything, it should be the other way around and update.sh should be changed to python3.

PR #1597, opened 2018-04-25
Option to check if owner is in a server and leave if they're not.

AleanaA:

----

### Description
Implements an option to check if the owner shares the servers with the bot. If the owner doesn't share the server with the bot, the bot will leave that server.

Credit to Winter for the idea and most of the code o//


musicbot/bot.py
```diff
@@ -939,6 +939,11 @@ def run(self):
             for s in self.guilds:
                 ser = ('{} (unavailable)'.format(s.name) if s.unavailable else s.name)
                 log.info(' - ' + ser)
+                if self.config.leavenonowners:
+                    check = s.get_member(owner.id)
```
jayktaylor: This may have an unintended result if the server is unavailable.

config/example_options.ini
```diff
@@ -144,6 +144,9 @@ ShowConfigOnLaunch = no
 # do not need to use "skip f" to force-skip a song, they will instead force-skip by default.
 LegacySkip = no
 
+# Leave servers if the owner is not found in them.
+LeaveNonOwners = no
```
jayktaylor: Maybe we can come up with a more descriptive name for this config option?

AutumnClove: I could understand that I just didn't want to make it to long, and keep it the same as the other config options. Will think of a better name.

musicbot/bot.py
AutumnClove: I'll add in a try and except?

jayktaylor: I have a suspicion that if a server is unavailable and you try to get a member, it'll return None regardless of if the member is in there or not because the data may not be obtainable. I'd rather err on the side of caution here and check to make sure that the server is available first.

config/example_options.ini
AutumnClove: How does LeaveNonBotOwners sound to you?

musicbot/bot.py
AutumnClove: Makes sense, I'll add in
```
If s.unavailable:
    unavailable_servers += 1
    log.info('Not going through with a check in {} servers due to unavailability.'format(unavailable_servers))
```
As well as a counter right below `if self.config.leavenonowners`

jayktaylor (changes requested):
Can you support strings from i18n files for the messages you send to the owner of the bot?

jayktaylor (approved):

PR #1654, opened 2018-07-03
Update to requirements.txt for docker

alucarddelta:

- [X] I have tested my changes on Python 3.5/3.6

----

### Description

PyNaCl is required to allow docker to work correctly. With out PyNaCl you get the error output below.

When doing `docker build .` the change will allow PyNaCl to be added, preventing the error.

Tested on ubuntu 16.04.4 LTS 

```
[ERROR:bot] Unknown error attempting to connect to voice
Traceback (most recent call last):
  File "/usr/src/musicbot/musicbot/bot.py", line 456, in get_voice_client
    vc = await self.join_voice_channel(channel)
  File "/usr/src/musicbot/musicbot/bot.py", line 416, in join_voice_channel
    voice = discord.VoiceClient(**kwargs)
  File "/usr/lib/python3.6/site-packages/discord/voice_client.py", line 217, in __init__
    raise RuntimeError("PyNaCl library needed in order to use voice")
RuntimeError: PyNaCl library needed in order to use voice
```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,3 +3,4 @@ pip
 youtube_dl
 colorlog
 cffi --only-binary all; sys_platform == 'win32'
+PyNaCl


AutumnClove:
Pr's go to review

AutumnClove:
Also for people who don't want to use docker this requires them to download something they don't need.  This still not might fix issues for some people since pynacl can be tricky to download http://i.imgur.com/5bQeXJX.png

alucarddelta:
If that is the case, maybe a change to the Dockerfile and adding a line to `&& pip install PyNaCl` after line 26. This way only people using docker will be affected by this change, and allow non docker users to be unaffected.

I have not tested that approach yet, but I do not see why it would not work.

AleanaA:
@AutumnClove libffi-dev and libsodium are both in the DockerFile already, so that won't be an issue.

@alucarddelta Although 16.04.4 is most likely the most widely used version of ubuntu at this point, I'd recommend attempting this on 14.04 and 18.04 as well, as 14.04 is still common, and 18.04 is the latest lts release.

Virtual machines do indeed work wonders for that kind of thing.

In any case, you should close this and make a pr for review instead, as it'll be closed anyways for this reason.

MattBSG:
The OS of the server the user is on should be irrelevant as Docker in of itself is more closely in-line with a VM as it will create an isolated environment with an image stated in the Dockerfile. If a Dockerfile works on one platform it **will** work on others as well (providing the version of docker is the same). I’ll test if this resolves it in a little bit and I’ll rebase + merge to review if I can reproduce.

MattBSG:
So after investigating your issue I have come to not be able to reproduce it on latest review. However, it does currently occur on master. Review is our next release and the already made fix will be merged eventually.

When you install pip modules, it checks if they have any dependencies then automatically installs those as well. `discord.py[voice]` already has `PyNaCl` as a dependency so there is no need to add it to the requirements; it will install on it's own. It is not `PyNaCl` that isn't being installed to begin with, it's that it likely installed wrong or was missing a dependency during build.

See these changes: 99cc25ac9ea386a958341654477b3c306dd40ba9

Issue #1678, opened 2018-08-01
BUG IN SETAVATAR


OzzyHelix:
Please tick all applicable boxes.

- [x] I am using Python 3.5 or higher (run `python --version` on the command line)
- [x] I have followed the [official guides](https://github.com/Just-Some-Bots/MusicBot/wiki) to install the bot for my system
- [x] I have updated my dependencies to the latest version using the appropriate `update` script

Which version are you using?

- [ ] The latest `master` version (1.9.7)
- [x] The latest `review` version

What type of issue are you creating?

- [x] Bug
- [ ] Feature request
- [ ] Question

----

### Description of issue
a bug in the setavatar command
there was someone who's case was not reproducible
in particular module 'aiohttp' has no attribute 'Timeout'
in the setavatar command changing
this
```
        try:
            with aiohttp.Timeout(10):
                async with self.aiosession.get(thing) as res:
                    await self.user.edit(avatar=await res.read())
```
to this
```
        try:
            with async_timeout.timeout(10):
                async with self.aiosession.get(thing) as res:
                    await self.user.edit(avatar=await res.read())
```
will fix the problem
however it might not be a long term solution however it does make the link uploading work at the very least 
### Steps to reproduce

1.  attempting to change the avatar from a link
note: uploading images as an avatar doesn't work because of this bug
----

### Log file
Please attach your MusicBot log file (located at `logs/musicbot.log`) to this issue. You can do so by dragging and dropping the file here. **If you do not include your log file, you WILL be asked to provide one.**
[musicbot.log](https://github.com/Just-Some-Bots/MusicBot/files/2248117/musicbot.log)

AleanaA:
I didn't test your solution, however 5 seconds in the aiohttp documentation, I found [this](http://docs.aiohttp.org/en/latest/client_quickstart.html#timeouts), and made 


        try:  
            timeout = aiohttp.ClientTimeout(total=10)  
            async with aiohttp.ClientSession(timeout=timeout):  
                async with self.aiosession.get(thing) as res:  
                    await self.user.edit(avatar=await res.read())
  
Although this works, it led to  

168118999337402368/Aleana#2643: !setavatar
[ERROR:bot] Exception in on_message
Traceback (most recent call last):
File "/mnt/secondary/aleana/test/MusicBot/musicbot/bot.py", line 2726, in on_message
    response = await handler(**handler_kwargs)
[...]

PR #1679, opened 2018-08-01
Fix setavatar for uploaded images and dev only commands

AleanaA:

----

### Description

[Fixes aiohttp error in setavatar](http://docs.aiohttp.org/en/latest/client_quickstart.html#timeouts) and updates attachments to work with rewrite

Edit: Also fixes a type error when running dev only commands

`TypeError: 'in <string>' requires string as left operand, not int`

#1678

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -2423,14 +2423,15 @@ class MusicBot(discord.Client):
         """
 
         if message.attachments:
-            thing = message.attachments[0]['url']
+            thing = message.attachments[0].url
         elif url:
             thing = url.strip('<>')
         else:
             raise exceptions.CommandError("You must provide a URL or attach a file.", expire_in=20)
 
         try:
-            with aiohttp.Timeout(10):
+            timeout = aiohttp.ClientTimeout(total=10)
+            async with aiohttp.ClientSession(timeout=timeout):
                 async with self.aiosession.get(thing) as res:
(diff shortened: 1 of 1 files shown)


musicbot/bot.py
```diff
@@ -2423,14 +2423,15 @@ def check(message):
...
+            thing = message.attachments[0].url
         elif url:
             thing = url.strip('<>')
         else:
             raise exceptions.CommandError("You must provide a URL or attach a file.", expire_in=20)
 
         try:
-            with aiohttp.Timeout(10):
+            timeout = aiohttp.ClientTimeout(total=10)
+            async with aiohttp.ClientSession(timeout=timeout):
```
jayktaylor: Why are you creating a new ClientSession here...

musicbot/bot.py
```diff
@@ -2423,14 +2423,15 @@ def check(message):
...
                 async with self.aiosession.get(thing) as res:
```
jayktaylor: ...when the next line uses a **different** ClientSession to make a call? You're wrapping one ClientSession's get request inside of another ClientSession? I saw on #1678 that you said you found http://docs.aiohttp.org/en/latest/client_quickstart.html#timeouts and added some code to fix it.

Then why aren't you just overriding it for this one `.get` call?

jayktaylor (changes requested):

jayktaylor (approved):

PR #1682, opened 2018-08-04
Fix unintended autoplaylist shuffle.

TheOneTwelfth:

----

### Description
Autoplaylist seems to be converting to set, therefore messing up the initial order even if random autoplaylist is turned off.


jayktaylor:
Not sure why this was a set, thanks.

PR #1708, opened 2018-08-26
Re-order and add 18.04 Ubuntu guide

AleanaA:

----

### Description
Re-orders the Ubuntu versions from 18.04 to 14.04 and adds an 18.04 guide.

### DIFF
--- a/_docs/installing/ubuntu.md
+++ b/_docs/installing/ubuntu.md
@@ -9,9 +9,31 @@ order: 1
 Installing MusicBot on Ubuntu via the command line is the **recommended way to install the bot**, though the system dependencies differ depending on what version of Ubuntu you are using. Firstly, lets install the dependencies required for your system:
 
 <details>
-  
-<summary><strong>Ubuntu 16.04</strong></summary>
+<summary><strong>Ubuntu 18.04</strong></summary>
+<br>
 
+~~~ bash
+# Install build tools
+sudo apt-get install build-essential unzip -y
+sudo apt-get install software-properties-common -y
+
+# Install system dependencies
+sudo apt-get update -y
+sudo apt-get install git ffmpeg libopus-dev libffi-dev libsodium-dev python3-pip 
+sudo apt-get upgrade -y
+
+# Clone the MusicBot to your home directory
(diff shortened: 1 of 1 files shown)


jayktaylor:
I'd usually merge this, but I'm just wondering... maybe there's a way we can make it so that we don't have to make a guide for every Ubuntu version? Or at least, a better way of having them on a page rather than having the instructions all crammed all over the place in `ubuntu.md`. Any ideas?

AleanaA:
I mentioned in the VIP channel on the help server that i wanted to use dropdowns instead to reduce the clutter, but codeblocks won't work in them on the site @ https://aleanaa.github.io/MusicBot/installing/ubuntu/

AleanaA:
@jaydenkieran Out of curiosity, is the 14.04 guide still needed? 14.04 is End of Life early next year, and I feel like just having guides for the latest two LTS releases in `ubuntu.md` would be fine.

MattBSG:
My two cents is no, it's not. However, as long as the bot *works* on 14.04 we should mark it as unsupported in the guides instead for those who may have legacy systems

AleanaA:
Hmm... So possibly a Legacy tab in the guides list? This would work for things such as CentOS as well for whenever CentOS 6 is EOL, and 8 is released.

jayktaylor:
+1 for marking 14.04 guide as unsupported/legacy

jogerj:
~~This won't work:
ppa:mc3man/xerus-media~~

~~Cannot add PPA: ''This PPA does not support bionic''.~~
> @jogerj The 18.04 guide I wrote doesn't add that ppa, or any for that matter.
> 
> This thread also isn't for support with the guides, it's for implementing an Ubuntu 18.04 guide, and moving or removing the 14.04 guide.

My bad, read the wrong lines.

Btw I'm running Linux Mint 19 (based on Ubuntu 18.04) I just used debian guide to run the bot and it works fine.

AleanaA:
@jogerj The 18.04 guide I wrote doesn't add that ppa, or any for that matter.

This thread also isn't for support with the guides, it's for implementing an Ubuntu 18.04 guide, and moving or removing the 14.04 guide.

PR #1736, opened 2018-09-17
add detail for updating from some of outdated version (guide)

TheerapakG:

----


AleanaA:
Hi there! Something mildly important as of 1.9.8 is that the bot no longer supports Python 3.5.2 and below. This is due to aiohttp being a dependency. I wouldn't generally make this distinction, but Ubuntu 14, 16, and Debian if I recall correctly, come shipped with 3.5.2, and it should be noted that they'll need one that's newer than that, instead of just stating that they'll need at least 3.5.

TheerapakG:
I made a change according to your comment @AleanaA

AleanaA:
Python 3.7 is supported, and is what's currently used to maintain the bot, along with being what's recommended in the windows guide. Support for 3.7 was brought in 1.9.8 as d.py rewrite fixed an issue relating to it. Versions prior to 1.9.8 will error if trying to use Python 3.7

TheerapakG:
@AleanaA Changed!
~(I somehow can't use it with 3.7 on 1.9.8, gonna try that again)~ Look like aiohttp stuff got fix I presume? I ran update script and run it a week ago and it break because something in aiohttp

MattBSG (changes requested):
Can you clean up the language and make it more professional fitting the style of our writing?

PR #1740, opened 2018-09-21
A Sloppy Permission Fix

### LINKED ISSUE #952: [Feature Request] Proper perms group for owner
Can we add a way of retaining ownership while toggling some permissions such as InstaSkip? I' d like to vote to skip instead of being a dictator. I very often find myself not skipping at all since I feel its unfair.

### LINKED ISSUE #1733: Owner being allowed to use play command etc
Please tick all applicable boxes.
- [x] I am using Python 3.5 or higher (run python --version on the command line)
- [x] I have followed the official guides to install the bot for my system
- [x] I have updated my dependencies to the latest version using the appropriate update script
Which version are you using?

- [x] The latest master version (1.9.8)

What type of issue are you creating?

- [ ] Bug
- [ ] Feature request
- [x] Question

----

### Description of issue
Hi, I've been using this bot for a very long time now ever since it came out but recently, the newer version does not allow owner of bot/server to use the !play command without having to be in the same voice channel. I remember the old version I was allowed to do this. Was this an intended change and is there a setting in either of the option.ini/permission.ini i can change to mimic how it was before? I would still prefer my regular members to have to be in the same voice channel but not the owner.

Thank you \o/

TheerapakG:

----

### Description
This is just a sloppy and partial fix for some ownerid related problems

close #1733, close #952

### DIFF
--- a/config/example_permissions.ini
+++ b/config/example_permissions.ini
@@ -82,6 +82,20 @@
 ;
 ;;;;;;;;;;;;;;;;;;;
 
+; This group is for owner.  Any options not specified will fallback to permissive default value.  Don't remove/rename this group.
+; You cannot assign users or roles to this group.  Those options are ignored.
+[Owner (auto)]
+; MaxSongLength = 0
+; MaxSongs = 0
+; MaxPlaylistLength = 0
+; AllowPlaylists = yes
+; InstaSkip = yes
+; Remove = yes
+; SkipWhenAbsent = no
+; BypassKaraokeMode = yes
+; ToggleAutoPlaylists = yes
+; Extractors = 
+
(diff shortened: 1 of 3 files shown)


TheerapakG:
As @AutumnClove suggested I added a new section, dedicated for the owner in the permission.ini file. These options, if not specified (which are not specified by default), will fallback to "permissive" settings.

AutumnClove:
Thanks, I just feel like having them setable allows flexibility for a per user basis

jayktaylor:
I'm merging this on the basis that it probably works and it's better than owner permissions being broken, but I'd like to actually clean it up at some point.

PR #1754, opened 2018-10-07
Added missing dependencies for CentOS7

jadissa:

----

### DIFF
--- a/update_dependencies_linux.sh
+++ b/update_dependencies_linux.sh
@@ -1,8 +1,9 @@
 #!/bin/bash
-
-python3.5 -V > /dev/null 2>&1 || {
-    echo >&2 "Python 3.5 doesn't seem to be installed.  Do you have a weird installation?"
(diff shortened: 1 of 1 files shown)


TheerapakG:
1. Please make a pull request targeting the `review` branch.
2. I also noticed that your repository is very out of date. The current version doesn't have the `update_dependencies_linux.sh`, only `update.sh`.
3. Last thing, the update script is used by multiple linux distro (not only CentOS) which might not have `yum` package manager.

AleanaA:
Putting in my two cents, you appear to have completely missed the guides. The script you changed is meant to run the bot, not update it, or any of the dependencies, and even then, there are no scripts that are meant to install the system dependencies, only ones through pip. All CentOS dependencies and a guide to installing the bot on CentOS can be found [here](https://just-some-bots.github.io/MusicBot/)

MattBSG:
The edited script is only for **updating** python dependencies. It is not mean't to be an install script nor would we want a CentOS only code that is for a script that is designed for all installations of Linux.

PR #1756, opened 2018-10-10
quick fix for leftover non-rewrite permissions

AleanaA:

----

### Description
Fixes
```client.py - line 221, in run_event 
    await coro(*args**kwargs)
boy.py - line 2600 in on_message
    user_permissions = self.permissions.for_user(mesage.author)
permissions - line 83, in for_user
    if user.id in group.user_list:
TypeError: `in <string>` requires string as a left operand, not int.
```
by turning `user.id` into a str, does the same for `role.id` in a later line.

### DIFF
--- a/musicbot/permissions.py
+++ b/musicbot/permissions.py
@@ -80,7 +80,7 @@ class Permissions:
         """
 
         for group in self.groups:
-            if user.id in group.user_list:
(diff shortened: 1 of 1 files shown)


AutumnClove:
Thanks bb

PR #1764, opened 2018-10-17
Use Python 3.6 with python3 in ubuntu guides

AleanaA:

----

### Description

Title is self-explanatory, `python3` shouldn't be Python 3.5.2 if there's a newer version being installed, fixes dependency issues relating to `run.sh` on first run.

`update.sh` will still output an aiohttp version issue as it uses `python3.5` instead of `python3` for some reason.

All Python 3.5.2 dependency related issues.


jayktaylor:
Removing a file from `/usr/bin` without good reason is almost always a bad idea. I don't think we should care if people have multiple Python versions. We're not really here to handhold.

PR #1776, opened 2018-10-29
load commands as cogs, playnext, data api, and various commands

TheerapakG:

- [x] I have tested my changes on Python 3.5/3.6 (webapi cog requires Python 3.7 only and this is one of the reasons that it is not loaded as default)

----

### Description
This is a `feature` PR.

1. Introduces cogs loader
> It's arguably more like extensions loader as it does not only load commands but can also run tasks on startup, forever (until cog got unloaded), and before unloading (cleanup)
2. Introduces command aliases (discard current aliases loading system because it's rather incompatible with the cog system)
3. Introduces a programmatic way to retrieve bot statuses (if enabled), to receive statuses you need to supply a command along with a correct permission token using HTTP/HTTPS protocols via port specified in the config file
4. Introduces replay command which adds the currently playing song to the end of the queue, or head of the queue if 'h' or 'head' command args given
5. Introduces playnext command
6. Introduces effect command that manages ffmpeg filters (with 'highly experimental' warning)
7. Introduces sudo command which executes musicbot command as other users

There are incentives for me to add commands to only this PR over than separating it to other PRs (I’m lazy)

List of new commands
> cogs-related
> - loadmodule
> - loadcog
> - unloadcog
> - cogmodule
> - (dev) cogops
> 
> alias-related
> - addalias
> - removealias
> 
> webapi-related (unloaded by default)
> - gentoken
> - revoketoken
> 
> queuemanip-related
> - playnext
> - replay
>
> playback-related
> - effect
>
> utility-related
> - sudo

Caution:
1. This will definitely break other PR adding commands.
2. This PR could introduce slightly longer startup time (~0 - 0.5s longer on my test machine, maybe significantly longer on a system with lower performance such as raspi). Running the bot after startup (1st run) seems to be somewhat slower than 2nd and 3rd consecutive run
3. This does not utilize `discord.ext.commands` (It's difficult to do so when we need to pass extra bot stuff such as player) hence the entire cog system is self-made, I will(?) try to migrate the cog system to utilize it in the upcoming PR(s?). This is a probably another PR that will happen in lots of phases.

overlapping features with #1547

This remove depending status of
https://github.com/TheerapakG/MusicBot/projects/1#card-16450708


TheerapakG:
Closing in favor of #1904

PR #1806, opened 2018-12-24
Typo from old version of guides.

AleanaA:

----

### Description

3.5 is no longer the version recommended by the guide, and as such, the add to path check box would not say to add 3.5 to path.


MattBSG (approved):
Good catch

Issue #1588, opened 2018-04-19
Command to check storage used.


SeasonalFerret:
Please tick all applicable boxes.

- [x] I am using Python 3.5 or higher (run `python --version` on the command line)
- [x] I have followed the [official guides](https://github.com/Just-Some-Bots/MusicBot/wiki) to install the bot for my system
- [x] I have updated my dependencies to the latest version using the appropriate `update` script

Which version are you using?

- [x] The latest `master` version (1.9.7)
- [ ] The latest `review` version

What type of issue are you creating?

- [ ] Bug
- [x] Feature request
- [ ] Question

----

### Feature request

I'm curious about how many videos my bot has downloaded and saved in audio_cache, how big the folder is, and how much room is left on my storage device.

I don't want to login to my VPS and turn off the bot just to check the file size, so it'd be super interesting to quarry the bot about it's environment.

Suggested/Example command would be !space.
Bot could respond with: "I have downloaded 23 songs for a total of 120MB. There is XGB left on the hard drive."
If integrated with #1584, it would say how much space is left that it's allowed to use instead of entire space left.

This command would simply be used for information/debugging, and could be used to get a kick out of if you have an overly insane amount of songs saved. Though, further information could possibly be giving to the user, such as total up-time, a ping test, average song duration, etc.

amit177:
Thank you, this will be added once the merge request will be finalized.

Issue #1584, opened 2018-04-17
audio_cache limit & auto-management.


SeasonalFerret:
Please tick all applicable boxes.

- [x] I am using Python 3.5 or higher (run `python --version` on the command line)
- [x] I have followed the [official guides](https://github.com/Just-Some-Bots/MusicBot/wiki) to install the bot for my system
- [x] I have updated my dependencies to the latest version using the appropriate `update` script

Which version are you using?

- [x] The latest `master` version (1.9.7)
- [ ] The latest `review` version

What type of issue are you creating?

- [ ] Bug
- [x] Feature request
- [ ] Question

----

### Description of issue

Set a storage limit for audio_cache. I don't have a 20TB harddrive, so limiting MusicBot to 1-2GB would be nice. A method for choosing what it should delete could be based off of what songs are worthless to keep.

Example: After audio_cache reaches (X)GB of data, go through and delete songs that either haven't been played in the past (X) weeks or that have only been played once or twice.

This would keep the audio_cache folder limited to only popular songs.

An exception may be songs on the autoplaylist.

PR #1884, opened 2019-04-23
Storagesize

johnathanyan:

----

### Description
This pull request targets #1584 and #1588. 

1. StorageLimit is a new configurable option in options.ini and the bot tracks the size of the audio cache after every download. I added another variable to config.py to track this field and also added this field to example_options.ini. If the storage size is exceeded, it deletes the least recently downloaded songs with the exception of songs on the queue and the song currently being played. It also takes into consideration if a single media file exceeds the entire storage limit. The functions used to calculate folder size and find the oldest media file are implemented in utils.py.

2. !space is a new command that prints out a message stating the size of the audio_cache folder and the storage size limit if specified.

#1584
#1588

### DIFF
--- a/config/example_options.ini
+++ b/config/example_options.ini
@@ -81,6 +81,10 @@ WhiteListCheck = no
 SkipsRequired = 4
 SkipRatio = 0.5
 
+# Specify a value in bytes to set a limit on the size of your audio cache. Once exceeded, it will
+# delete the oldest songs that are cached. If left blank, there is no limit to the size of your cache.
+StorageLimit =
+
 # Determines if downloaded videos will be saved to the audio_cache folder. If this is yes,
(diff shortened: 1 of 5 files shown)


AutumnClove (changes requested):
Overall a nice pull request that adds a nice feature. Me and @TheerapakG had a discussion about something though:
No need for 500+ os.path.getsize calls every time we delete a file if the user has a large audio_cache or something. Uninteded long processing times could occur. I'd like to see an implementation of a check of how much we need to delete, and delete files until we reach the needed amount. 
https://github.com/Just-Some-Bots/MusicBot/pull/1884/files#diff-29f8598df1d9c6ef8ce6e87255a1c64cR331

AutumnClove:
Closes for inactivity

PR #1839, opened 2019-02-09
playground_defrag: group together relevant methods into classes and reduce length in long file [1]

TheerapakG:
After creating your pull request, tick these boxes if they are applicable to you.

- [x] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [x] I have tested my changes on Python 3.5/3.6

----

### Description
[1]: guild, vc, message methods grouping
[2]: move toward using contexts -> depends on cog
[3]: move logging stuff into its own module
[4]: move command handling to ManagedGuild object
[5]: prettify calling managed stuff

Group together relevant methods into classes and reduce the length in long files (separation of concerns).
This PR, combined with cog PR, would reduce loc in bot.py by ~ 2000 lines. 
Part of rw.
Next phases of these PRs is paused in development waiting for the adoption of the cog system because these two PRs will result in a huge merge conflict.

If the bot is closed via the x button on Windows (the way that is not preferred), you have to wait for Discord to "realize" that the bot got disconnected before starting the bot up again or else the bot will not output any sound. However, KeyboardInterrupt can be used to stop the bot cleanly.

If both #1776 and this will both get merge, merge #1776 first and then merge [TheerapakG/MusicBot:playground_cogs_defrag](/TheerapakG/MusicBot/tree/playground_cogs_defrag) (I will merge it into this branch when that time came)

If #1776, #1727 and this will get merged together, merge #1727, #1776 (after cog_autostream merged into it) and then this [TheerapakG/MusicBot:playground_autostream_cogs_defrag](/TheerapakG/MusicBot/tree/playground_autostream_cogs_defrag) (I will merge it into this branch when that time came)

PR #1904, opened 2019-05-14
commands as cog v2

TheerapakG:

In no way the test I did were a thorough test.

----

### Description
Doing what #1776 and #1839 does but in a discord.ext.commands and more sensical way.

This PR changes several models related to player and playlists for ease of usage in the future; hence, data saved in the data folder will be incompatible. There are checks for the version of the data file to prevents incompatible saves.

Player got reworked and have some states removed. Check the implementation for details.

If there is not any entry in the playlist, player would now poll the playlist at an interval for next entry rather than stopping itself.

Playlists got reworked and now do not take raw urls and make entries inside itself. It relies on entries added to it instead. It can also be named.

Playlists are now objects which are in control of caching instead and hence, can be set to cache multiple entries in the list before player pull it out. Check the implementation for details.

Downloader got reworked and now use its internal loop which it got when it is instantiated instead. Check the implementation for details.

Entries have dedicated function to create them. Check the implementation for details.

Due to the implementation of the cog loader (copied from ModuBot), run script is modified.


TheerapakG:
Seems to be considerably stable now so I would work on embeds next.

TheerapakG:
Embeds works are done or at least I think so. Untested though.

PR #1727, opened 2018-09-12
Playlist Functionalities [1]: Autostream Implementation

TheerapakG:
After creating your pull request, tick these boxes if they are applicable to you.

- [x] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [x] I have tested my changes on Python 3.5/3.6

----

### Description
This is a `feature` PR.

> I planned to do several pull requests to rewrite and add functionalities to autoplaylist which cover a range of feature requests. This will happen in different phases. Phases are arranged in the order that it won't interfere with each other very much.
> 
> phase 1: add autostream and related functionalities
> phase 2: multiple playlists, make options to allow totally different autoplaylist.txt in different servers (or guild)
> phase 3: add an option to show autoplaylist "now playing" notifications

This pull request introduces autostream. It's similar to autoplaylist, but it's for streams.
You can add streams for autoplaying in config/autostream.txt.
If the configuration for randomization and toggling is not set, autostream will play after autoplaylist.
Autostream and autoplaylist entries could be set to skip automatically when someone adds stuff to the queue.

Autostream and autoplaylist can play in 2 modes:
- merge: merge autoplaylist and autostream when playback
- toggle: use the `toggleplaylist` command to toggle between them

- modes are saved
- different modes for different servers possible

> my cog PR break this. There is alternative cog implementation of this PR in my [autosteam_cogs](/TheerapakG/MusicBot/tree/autosteam_cogs) branch. I will merge it into this branch when the cog PR got merged

This remove depending status of
https://github.com/TheerapakG/MusicBot/projects/1#card-16450669

### Related issues (if applicable)
(close #1462, close #1587)
not a direct implementation of #1587 but introduce "roughly equivalent" features, here is the breakdown:
1. autostream.txt now exist
2. you can add or remove a stream using `autostream` command
3. toggling can be set using the config file
4. state of toggling does save so next startup it'll be on whatever mode you've set
5. you can use the config file to skip the autos when added the new entry (the self-closing stream do exist so I added the config to let you choose to force the autostream to skip or not)

PR #1911, opened 2019-05-27
[CogV2] Playlist Functionalities [1]: Autostream Implementation that works with #1904

TheerapakG:

----

### Description
excerpt from #1727:
> I planned to do several pull requests to rewrite and add functionalities to autoplaylist which cover a range of feature requests. This will happen in different phases. Phases are arranged in the order that it won't interfere with each other very much.
> 
> phase 1: add autostream and related functionalities
> phase 2: multiple playlists, make options to allow totally different autoplaylist.txt in different servers (or guild)
> phase 3: add an option to show autoplaylist "now playing" notifications

This pull request introduces autostream. It's similar to autoplaylist, but it's for streams.
You can add streams for autoplaying in config/autostream.txt.
If the configuration for randomization and toggling is not set, autostream will play after autoplaylist.
Autostream and autoplaylist entries could be set to skip automatically when someone adds stuff to the queue.

Autostream and autoplaylist can play in 2 modes:
- merge: merge autoplaylist and autostream when playback
- toggle: use the `toggleplaylist` command to toggle between them

This implementation is different from the one in #1727 as this one implement only the basic features and lack the ability to save modes as the playlist system will have to be revise in the later date. for multiplaylists anyway


TheerapakG:
Closing this to allow more streamlined bug fixing commits merging as #1917 is gradually getting "rather stable" and is also designed to supersede this pull request.

Issue #1650, opened 2018-06-27
Change Footer Text


timbirtles:
Please tick all applicable boxes.

- [Y ] I am using Python 3.5 or higher (run `python --version` on the command line)
- [Y] I have followed the [official guides](https://github.com/Just-Some-Bots/MusicBot/wiki) to install the bot for my system
- [Y ] I have updated my dependencies to the latest version using the appropriate `update` script

Which version are you using?

- [Y ] The latest `master` version (1.9.7)
- [ ] The latest `review` version

What type of issue are you creating?

- [ ] Bug
- [ ] Feature request
- [Y] Question

----

### Description of issue
Is it possible to change the footer of embeds to say what we like instead of 'Just-Some-Bots/MusicBot (1.9.7)'

### Steps to reproduce

1. 
type a command
![image](https://user-images.githubusercontent.com/33937835/41957595-618710dc-79df-11e8-8617-658ee15c47c8.png)

----

### Log file
Please attach your MusicBot log file (located at `logs/musicbot.log`) to this issue. You can do so by dragging and dropping the file here. **If you do not include your log file, you WILL be asked to provide one.**
`[66.876649857] 2018-06-27 07:45:56,948 - INFO - launcher: Starting sanity checks
[67.871809006] 2018-06-27 07:45:56,949 - INFO - launcher: Checking for Python 3.5+
[69.868564606] 2018-06-27 07:45:56,951 - INFO - launcher: Checking console encoding
[70.367813110] 2018-06-27 07:45:56,952 - INFO - launcher: Setting console encoding to UTF-8
[70.367813110] 2018-06-27 07:45:56,952 - INFO - launcher: Ensuring we're in the right environment
[71.865320206] 2018-06-27 07:45:56,953 - INFO - launcher: Adding local bins/ folder to path
[72.865009308] 2018-06-27 07:45:56,954 - INFO - launcher: Required checks passed.
[75.359582901] 2018-06-27 07:45:56,957 - INFO - launcher: Optional checks passed.
[81.846237183] 2018-06-27 07:45:56,963 - INFO - launcher: Moving old musicbot log

######################### PRE-RUN SANITY CHECKS PASSED #########################

[1.6604070663452148] 2018-06-27 07:45:58,542 - WARNING - musicbot.config | In config.py::MainThread(7944), line 112 in run_checks: i18n file does not exist. Trying to fallback to config/i18n/en.json.
[...]

PR #1922, opened 2019-06-18
Improvement add custom embed footer

AutumnClove:

----

### Description
A lot of people ask about removing the footer, so this pr adds a command to provide the same info and a little extra info. As such there's now an option to disable the footer. 
This can be furthered for custom footers but I decided against doing that since some footer stuff needs users to know coding such as icon_url and what not.
Thanks @TheerapakG for bug fixing my typo

#1819
#1650


jayktaylor (changes requested):
The info command feels kinda pointless and if people want to remove the footer they can modify the code. The absolute maximum I'd be happy with is allowing custom configurable embed footers in the config file, even though - again - people can just edit the bot file.

musicbot/bot.py
```diff
@@ -1109,9 +1109,15 @@ def run(self):
 
     def _gen_embed(self):
         """Provides a basic template for embeds"""
+        if not self.config.footer_text:
+            text = 'Just-Some-Bots/MusicBot ({})'.format(BOTVERSION)
```
jayktaylor: Why not just set the default in config.py to this rather than doing this conditional?

musicbot/config.py
```diff
@@ -355,6 +357,8 @@ class ConfigDefaults:
     legacy_skip = False
     leavenonowners = False
     usealias = True
+    disableembedfooter = True
```
jayktaylor: The default for this should be `False`

jayktaylor (changes requested):

musicbot/config.py
AutumnClove: since we're setting the default footer in config.py there's really no need for this option anymore so I just ended up removing this

musicbot/bot.py
AutumnClove: done

jayktaylor (approved):

PR #1945, opened 2019-09-02
Fix legacyskip

AutumnClove:

- [x] @Roadhog360  and I have tested my changes on Python 3.5.3 or higher

----

### Description
Fix for legacyskip changes suggested by @TheerapakG
Issue bought up by @Roadhog360

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -2645,7 +2645,7 @@ class MusicBot(discord.Client):
         if message.author.bot and message.author.id not in self.config.bot_exception_ids:
             log.warning("Ignoring command from other bot ({})".format(message.content))
             return
-            
+
         if (not isinstance(message.channel, discord.abc.GuildChannel)) and (not isinstance(message.channel, discord.abc.PrivateChannel)):
             return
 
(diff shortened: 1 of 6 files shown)


TheerapakG:
That doesn't seem right. Why check if both forced and legacy?

AutumnClove:
@TheerapakG doing and not doesn't let users force skip still and vote skip, doing or just raises perms error, doing this allows users to do !skip f with legacy skip on and force skip or do

musicbot/bot.py
```diff
@@ -2088,7 +2088,7 @@ def check(reaction, user):
                 player.skip()  # TODO: check autopause stuff here
                 await self._manual_delete_check(message)
                 return Response(self.str.get('cmd-skip-force', 'Force skipped `{}`.').format(current_entry.title), reply=True, delete_after=30)
-            else:
+            if(param.lower() in ['force', 'f']):
```
jayktaylor: This logic is messy and should be rewritten rather than repeating code.

jayktaylor (changes requested):

musicbot/bot.py
TheerapakG: Does the rewrite acceptable?

jayktaylor (approved):
That's cleaner, thanks

PR #1946, opened 2019-09-10
Fix for showing negative estimated time for entries w/o duration information from ytdl

TheerapakG:

----

### Description
Warn no duration data instead of showing negative estimated time. This also adds ffprobe-ing or use pymediainfo if available to get the duration of downloaded entries.

### DIFF
--- a/musicbot/entry.py
+++ b/musicbot/entry.py
@@ -81,13 +81,17 @@ class BasePlaylistEntry(Serializable):
 
 
 class URLPlaylistEntry(BasePlaylistEntry):
-    def __init__(self, playlist, url, title, duration=0, expected_filename=None, **meta):
+    def __init__(self, playlist, url, title, duration=None, expected_filename=None, **meta):
         super().__init__()
 
         self.playlist = playlist
         self.url = url
         self.title = title
         self.duration = duration
+        log.info('Cannot extract duration of the entry. This does not affect the ability of the bot.'
+                 'However, estimated time for this entry will not be unavailable and estimated time'
+                 'of the queue will also not be available until this entry got downloaded.'
+                 'entry name: {}'.format(self.title))
         self.expected_filename = expected_filename
         self.meta = meta
         self.aoptions = '-vn'
@@ -319,6 +323,29 @@ class URLPlaylistEntry(BasePlaylistEntry):
                 # Move the temporary file to it's final location.
                 os.rename(unhashed_fname, self.filename)
 
+        # Get duration from the file after downloaded
+        args = [
+            'ffprobe', 
+            '-i', self.filename, 
+            '-show_entries', 'format=duration', 
+            '-v', 'quiet', 
+            '-of', 'csv="p=0"'
+        ]
+
+        output = await self.run_command(args)
+        output = output.decode("utf-8")
+
+        try:
+            self.duration = float(output)
+        except ValueError:
+            # @TheerapakG: If somehow it is not string of float
+            if not self.duration:
+                log.error('Cannot extract duration of downloaded entry, invalid output from ffprobe.'
+                          'This does not affect the ability of the bot. However, estimated time for this entry'
+                          'will not be unavailable and estimated time of the queue will also not be available'
+                          'until this entry got removed.'
+                          'entry file: {}'.format(self.filename))
+
 
 class StreamPlaylistEntry(BasePlaylistEntry):
     def __init__(self, playlist, url, title, *, destination=None, **meta):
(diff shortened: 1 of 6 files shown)


musicbot/entry.py
```diff
@@ -319,6 +324,29 @@ def is_exe(fpath):
...
+            self.duration = float(output)
+        except ValueError:
+            # @TheerapakG: If somehow it is not string of float
+            if not self.duration:
+                log.error('Cannot extract duration of downloaded entry, invalid output from ffprobe.'
+                          'This does not affect the ability of the bot. However, estimated time for this entry'
+                          'will not be unavailable and estimated time of the queue will also not be available'
+                          'until this entry got removed.'
+                          'entry file: {}'.format(self.filename))
+
```
srhinos: This is an expensive thing to run every time for every song, maybe we should only do this if the duration is None?

musicbot/entry.py
```diff
@@ -328,7 +356,10 @@ def __init__(self, playlist, url, title, *, destination=None, **meta):
         self.url = url
         self.title = title
         self.destination = destination
-        self.duration = 0
+        self.duration = None
+        log.info('Cannot estimate duration for stream, estimated time of the queue will not be available'
+                 'until this stream got removed from the queue.'
+                 'stream name: {}'.format(self.title))
```
srhinos: We didn't log this before, no reason to log it now.

musicbot/playlist.py
```diff
@@ -344,11 +344,17 @@ def peek(self):
         """
             (very) Roughly estimates the time till the queue will 'position'
         """
-        estimated_time = sum(e.duration for e in islice(self.entries, position - 1))
+        if any(e.duration for e in islice(self.entries, position - 1) == None):
+            raise InvalidDataError('no duration data')
```
srhinos, line 348: We shouldn't return an error if no duration is found. Would be much more user friendly to return a string stating that no duration is found because its not an error, its an expected condition.

musicbot/playlist.py
```diff
@@ -344,11 +344,17 @@ def peek(self):
...
+        else:
+            estimated_time = sum(e.duration for e in islice(self.entries, position - 1))
 
         # When the player plays a song, it eats the first playlist item, so we just have to add the time back
         if not player.is_stopped and player.current_entry:
-            estimated_time += player.current_entry.duration - player.progress
+            if player.current_entry.duration == None: # duration can be 0
+                raise InvalidDataError('no duration data in current entry')
```
srhinos, line 355: We shouldn't return an error if no duration is found. Would be much more user friendly to return a string stating that no duration is found because its not an error, its an expected condition.

srhinos (changes requested):

musicbot/playlist.py
TheerapakG: I'd argue that raising an error here is better than returning a string and then the caller has to check for type returned to format accordingly. Hence, I modified the code in the play command to not raise an error but to change the reply message instead.

musicbot/entry.py
TheerapakG: If you say so then I will do it that way. I was just being a little pedantic that if some commands show that duration cannot be estimated and there's nothing mentioned then some users might come to ask us why.

musicbot/playlist.py
srhinos: Maybe make this a bit more wordy then? Less than a three word error so that way a user understands the bot isn’t broken from this error and they did nothing wrong, they just queued a song we couldn’t fetch the time info for

srhinos: Ah, i see you’re handing this in the create embed. I must have missed this the first time.

musicbot/entry.py
srhinos: Is there no other way to fetch the duration of the file than to do this? I’d be fine with something that didn’t rely on running an ffmpeg command on the cli and will be fine with approving this as it’ll rarely activate but maybe looking into a less hacky solution now will do us better for preventing future technological debt.

srhinos (dismissed):
Check into the reopened comment and if there’s no better solution, ask for re-review and I’ll approve

musicbot/entry.py
TheerapakG: Does this suffice an additional pip requirement?

srhinos: I'd prefer not to tbh if it can be avoided

TheerapakG: I'm afraid that that is impossible with something we currently have in requirements. This would need something like pymediainfo.

AleanaA (approved):

PR #1959, opened 2019-10-14
Autoinstaller + service file and command

AleanaA:

----

### Description
Creates an installer script that users can pull and easily install the bot with a single command
Adds a service file to integrate the bot with systemctl to allow for launching on startup and restarting on crash
Adds a command file that allows for easier management of the bot through systemctl
Command information can be found by running `musicbot` or `musicbot --help`

### DIFF
--- a/musicbot.service
+++ b/musicbot.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=MusicBot
+
+[Service]
+WorkingDirectory=mbdirectory
+ExecStart=/usr/bin/pythonversionnum mbdirectory/run.py
+Restart=on-failure
+
+[Install]
+WantedBy=default.target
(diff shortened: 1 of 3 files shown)


musicbot.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=MusicBot
+
+[Service]
+WorkingDirectory=mbdirectory
+ExecStart=/usr/bin/pythonversionnum mbdirectory/run.py
```
srhinos, line 6: does `pythonversionnum` actually work when pulling the latest python version?

AleanaA:
It's replaced by the install script in the same way config files are
managed, just using the gathered py version instead of user input.

srhinos:
oh nice okay

srhinos (approved):

PR #1970, opened 2019-10-22
Normalize audio with loudnorm instead of volumedetect.

SMarioMan:
The current volumedetect approach measures the loudest part of the file and increases the gain until it's just shy of clipping. This is a good first approach, but a bit naive for actually targeting a certain loudness. More dynamic tracks will still seem much quieter than tracks continually pushing the loudness ceiling. 
loudnorm takes more time to process audio than volumedetect, but in return will properly normalize to a target loudness using the EBU R128 loudness standard. See: http://k.ylo.ph/2016/04/04/loudnorm.html

### DIFF
--- a/musicbot/entry.py
+++ b/musicbot/entry.py
@@ -218,8 +218,7 @@ class URLPlaylistEntry(BasePlaylistEntry):
 
             if self.playlist.bot.config.use_experimental_equalization:
                 try:
-                    mean, maximum = await self.get_mean_volume(self.filename)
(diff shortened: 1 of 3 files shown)


srhinos:
Oh yeah, I coded that and it was really bad. I was just upset with the random jumps in loudness and so I wanted to push low volume music up and push high volume music down.

I can test that this still works later tomorrow since just reading it, it seems fine unless someone else wants to trial it and attest to its function

TheerapakG (approved):
Works fine for me

jayktaylor (approved):

PR #1971, opened 2019-10-24
fix check for python3 to not only check for 3.5

AleanaA:

----

### Description
Checks if python3 version is 5, and if not, also actually checks for version 6 or higher, as it was supposed to in the first place.


TheerapakG (commented):
doesn't [this](https://github.com/Just-Some-Bots/MusicBot/pull/1971/files#diff-1b0c2b516b83393edb7200ad5ff12181R37-R40) already do that?

PR #1975, opened 2019-10-27
remove `'source_address': '0.0.0.0'` from musicbot/downloader.py since it causes 'Unable to extract video data' error in ytdl

infowolfe:

- [X] I have tested my changes on Python 3.5.3 or higher

----

### Description

This PR fixes 
```
ERROR: xSaPCoTZr-8: YouTube said: Unable to extract video data
```

There's a few, mostly marked (incorrectly) as duplicate in the last couple weeks.

Also, I'll take an unban forthwith from the rhinobot discord as infowolfe#1337 and an apology from the "helper" who said that 1) my bug isn't a bug and 2) that they tested on centos and that means that the problem is one from my environment.

### DIFF
--- a/musicbot/downloader.py
+++ b/musicbot/downloader.py
@@ -19,7 +19,6 @@ ytdl_format_options = {
     'quiet': True,
     'no_warnings': True,
     'default_search': 'auto',
-    'source_address': '0.0.0.0',
     'usenetrc': True
 }
 


amit177:
Do you have more than one IP on your machine?

infowolfe:
Do you not? It's a linux box... I have lo (loopback). I have a docker bridge. I have a virbr0 for virt-manager. I have tunnel interfaces for openvpn termination. I have ipv4/ipv6 dual stack.

amit177:
I'm not talking about loopback, I'm talking about IPs you access the network with.
The only thing that line does is tell python to use any of the IPv4s in the machine.
Do you have more than one public IPv4?

infowolfe:
I have 0 public ipv4 ips on this machine, it's using RFC1918 space only (for ipv4). There's a single default route and a single private ipv4 IP on enp1s0. I *do* have ipv6 also, but 0.0.0.0 is ipv4 only, not the ipv6 equivalent `::`.

amit177:
and on the gateway, do you have 1x ipv4 and 1x ipv6 (public)? or just 1x ipv4?

infowolfe:
As is typical in a native dual-stack deployment I have a _block_ of ipv6 ips that're being advertised through ipv6 RA's on the lan. I have a single ipv4 address assigned to the gateway from my isp.

**But** ipv6 shouldn't really even matter. If you're selecting an ipv4 address as your source address, then the traffic shouldn't touch the ipv6 network at all. Regardless, I reported the bug with `--source-address` to upstream. This is a fix for people who're having this specific issue, of which there have been a few, if the recently closed issues are an indicator.

Roadhog360:
What I think is happening then is the bot is using the ipv4 address, which could have been blocked by YouTube, but when you try the youtube-dl command, it will work fine since it might be using one of the ipv6 addresses, which aren't blocked, hence why you'd get errors when using the bot but not the youtube-dl command.
This isn't a bug and is an ongoing issue with YouTube battling music bots and ripping services, blocking anything it deems not an actual viewer.

infowolfe:
You sure about that? Shouldn't this fail then?
```
freya:/dev/shm$ curl -4 "https://r2---sn-bvvbax4pcxg-gq0e.googlevideo.com/videoplayback?expire=1572243386&ei=WjO2Xcb0GrenhwaQwICIDg&ip=2605%3Aa601%3Aa608%3Acc00%3A42a8%3Af0ff%3Afe3d%3A2c7b&id=o-AC5VXhUX73D4oTP3SoCVGj6Lb0qgblGvpfdrBWK7JcEu&itag=303&aitags=133%2C134%2C135%2C136%2C137%2C160%2C242%2C243%2C244%2C247%2C248%2C278%2C298%2C299%2C302%2C303&source=youtube&requiressl=yes&mm=31%2C29&mn=sn-bvvbax4pcxg-gq0e%2Csn-p5qs7n7l&ms=au%2Crdu&mv=m&mvi=1&pl=42&initcwndbps=2728750&mime=video%2Fwebm&gir=yes&clen=81951911&dur=359.933&lmt=1540318494342914&mt=1572221635&fvip=5&keepalive=yes&fexp=23842630&c=WEB&txp=5432432&sparams=expire%2Cei%2Cip%2Cid%2Caitags%2Csource%2Crequiressl%2Cmime%2Cgir%2Cclen%2Cdur%2Clmt&lsparams=mm%2Cmn%2Cms%2Cmv%2Cmvi%2Cpl%2Cinitcwndbps&lsig=AHylml4wRAIgLP6YEoRtpmgO9ERi5nV55ffinz1RcD7C_yHPoFTEOZ0CID3Q56M1s00VX-tJpWaE3CI2JpYk8AvYMa2UjRH3iDd_&sig=ALgxI2wwRAIgBE7307Kn4dDg3o_6dMhgw7bdSAetyC9blW7BQOuHa5QCIDy-QXaGmA9EQREq_KsPEQW2F9SNFNzZtd1l-ELvvLUA&ratebypass=yes" -o testing.file
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
 14 78.1M   14 11.7M    0     0   972k      0  0:01:22  0:00:12  0:01:10  441k
```

infowolfe:
@Roadhog360 you're probably right. I disabled ipv6 completely (`sysctl -w net.ipv6.conf.all.disable_ipv6=1`) and attempted without a `--source-address` and it reacted the same way as it was acting when I had one specified. I wonder why the `curl` worked fine then while also limited to ipv4-only. Either way, it's easy to switch ipv6 ips, so generally speaking unless that 4 year old ipv6 bug is still a bug (and it doesn't seem to be, at least on my system), you might consider changing that `0.0.0.0` to a `::`. I tested with ipv6 enabled using `youtube-dl -4` and `youtube-dl -6` and got the same behavior as the ipv4 `--source-address` vs unspecified... so apparently youtube don't like my ipv4 ip in combination with `youtube-dl`'s access patterns.

amit177:
The reason IPv4 is being forced is that in 2019, there are still websites that do not support IPv6.
Youtube does, that's correct (which as you said in your explanation, that was your solution), but Youtube is not the only website you can download from.
We don't want to allow IPv6 by default for that reason, it will be enabled once it starts being more widespread.
Thank you for testing it on your end.

jayktaylor:
Why don’t we just make this an advanced config option if necessary? This line was added to the file [a few years ago](https://github.com/Just-Some-Bots/MusicBot/commit/8963cf463c08653d99fd53502af495193164039c) because of issues with IPv6 (to be fair, the support for IPv6 then was probably less than now).

I don’t think it’s unreasonable to allow advanced users to mess with yt-dl options like this if they know what they’re doing and think it’ll make a difference. Obviously they *could* just modify the code, but that’s more annoying than just changing a config file line.

amit177:
Sounds good to me

PR #1978, opened 2019-10-29
Get Token from os.environ

ghostrick:

----

### Description
Changed to get token from environment variable.
if not present, Read from options.ini,


amit177:
I don't see any requests regarding this change nor a use case

PR #1980, opened 2019-11-05
Travis CI:  Upgrade from Python 3.5 to 3.8

cclauss:

----


TheerapakG:
We need to check whether 3.5 would work as we officially supported 3.5 so I don't think we should do this. If 3.5 works then 3.8 should work (with exception of the logging format thing that it now checks by default)

jayktaylor (approved):

PR #1991, opened 2019-11-19
Serbian language added

DedaDev:

----


AleanaA (changes requested):
This PR would delete the example_aliases.ini and example_options.ini files, which should not be done. It would also change strings that are hard-coded and modify requirements.txt. These changes should be listed in the description, and even then, aren't fit for this PR.

amit177:
The required changes were not completed.

PR #2010, opened 2020-02-08
Set now playing as channel topic

chucklesb:

----

### Description
Added optional feature to display the now playing title in the channel topic.


musicbot/bot.py
```diff
@@ -662,6 +679,19 @@ def _autopause(player):
                 await self.change_presence(activity=game)
                 self.last_status = game
 
+    async def update_now_playing_topic(self, player, entry=None, is_paused=False):
+        guild = player.voice_client.guild
+        last_np_msg = self.server_specific_data[guild]['last_np_msg']
+        channel = last_np_msg.channel if last_np_msg else None
```
AleanaA: It would likely be better if there was an additional option to specify a channel instead of using the last now playing message.

chucklesb: You're right, it is implemented now.

musicbot/bot.py
```diff
@@ -662,6 +679,16 @@ def _autopause(player):
...
+    async def update_now_playing_topic(self, entry=None, is_paused=False):
+        for id in self.config.nowplaying_topic_channels:
+            channel = self.get_channel(id)
+            if entry and is_paused:
+                await channel.edit(topic=':pause_button: **' + entry.title + '**')
+            elif entry and not is_paused:
+                await channel.edit(topic=':arrow_forward: **' + entry.title + '**')
+            elif entry is None:
+                await channel.edit(topic=self.config.default_topic)
+
```
amit177: Could you please add a permission check? so it doesn't just throw an exception if the bot doesn't have Manage Channels.
If it doesn't have permission, log a warning. otherwise, continue to the rest of the function.

amit177 (changes requested):

jayktaylor (changes requested):
Does not account for the bot playing music in multiple servers

PR #2014, opened 2020-03-01
Update Dockerfile

kvdheijden:

----

### Description
Update the Dockerfile to work with an Alpine linux v3.11 base image.

The Dockerfile is using python3.8.2-r0. This can be changed to 3.7.5-r1 by using Alpine linux v3.10 as base (the current master branch doesn't work with python 3.8).

The Dockerfile is also restructured to allow better caching when the requirements.txt is not updated.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,4 @@
-FROM alpine:edge
-
-# Add project source
-WORKDIR /usr/src/musicbot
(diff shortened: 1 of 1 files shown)


AleanaA (approved):

TheerapakG (approved):

amit177:
Thank you for the submission

PR #2030, opened 2020-03-18
cogs-rewrite update tracking review/master branch after release-260819 (iteration 3)

TheerapakG:

----

## Description
This PR either update cogs-rewrite branch to be up to date against the review branch, changes I made to the cogs-rewrite in the meanwhile, or both.
#### This is expected to be merged into cogs-rewrite at most not long after the review branch is merged into master or when diff exceeds the tolerable limit.

Extending from the above description, this PR also aims to make the bot more modular in various regards.
- Introducing "entry builders" for different types of entries as opposed to hard-coding the logic inside the play command, effectively reducing the complexity of the function.
- Decoupling the logic of MusicBot from normal stuff that every bot has. In this PR, player, playlist, and autoplaylist logic are now nearly 100% outside of `SmartGuild` (earlier named `RichGuild`).

### Data incompatibility
This PR cannot load data saved by the current cogs-rewrite version. You're advised to back up those data before trying out this PR.

### Update Against the Review Branch:
- #2025 "Fix disconnect bug"
- #2033 "Resolve issues created by last most recent PR"

#### Tested
- broken up the `async QueueMagement._play` into several "entry builders" as I feel that the method has become a "god method" or close if it's not.
    - I want to do this for a long time but did not have time to get to it and the PR was so huge already.
- `play` command and its variants currently allows stream URLs
- `fromfile` command now does not take `mode` argument.
- `fromfile` command now takes .txt file containing any combination of local path, Spotify URIs, URLs that YT-DL supports, Youtube queries and stream URLs. Previously it only takes .txt file containing only URLs that YT-DL supports in the playlist mode and only stream URLs in the stream mode.
- changes deserialization of guild.
- various things now use rlock instead of asyncio.lock where appropriate (anticipating multithreading... or not)

#### Untested
- changes smart_guild and player structure
- changes smart_guild deserialization
- added `crossmodule.export` wrapper and related utilities to make exporting and integrating between functions between different modules easier

### Bugfixes for cogs-rewrite:
- change `self.str` to `ctx.bot.str` when getting string for the reply text due to changes in the bot structure. This was likely forgotten by me.
- `async playlist.add_entry` now return position like expected


TheerapakG:
Finally, after force-pushing the check runs. Not sure what happened before.

PR #2055, opened 2020-04-19
Fix some oversights in cogs rewrite branch

jokey2k:

- [X] I have tested my changes on Python 3.5.3 or higher

----

### Description

While running the bot, it crashed at various positions


TheerapakG:
Thank you for your contribution! However, cogs-rewrite is still a work-in-progress/prototype thing (there is a reason why it isn't going into the review branch directly) with things like #2030 that already solve some of this issue and would likely create more issue (more so when the last commit is expected to broken a lot of things). I don't think that improving stability is a goal at this point when code is still moving around a lot.

PR #2061, opened 2020-04-25
Update Dockerfile

andrew000:
Works good on ubuntu:18.04 and python3.6

- [+] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [+] I have tested my changes on Python 3.5.3 or higher

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,42 +1,24 @@
-FROM alpine:3.11
+FROM ubuntu:18.04
+
+# Add project source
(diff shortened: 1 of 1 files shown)


srhinos:
We dont want to fundamentally change the OS

andrew000:
> We dont want to fundamentally change the OS

Check your Dockerfile.

It doesn't work with apline

srhinos:
we merged in a better fix for docker around when I originally commented here

PR #2076, opened 2020-05-24
Update deps version

amit177:

----

### Description
Remove version requirements from dependencies


AleanaA (approved):
If we're going this route, as d.py includes pynacl as a requirement and works fine presently, we could probably just remove pynacl from requirements.txt as a whole

PR #2166, opened 2020-11-08
Update requirements.txt

jamesonp:
Fixes dependencies required to make bot run


AleanaA:
Pull requests should target review. We have a template that says this.

Issue #2176, opened 2020-12-21
Error updating bot


tommyruen:
Please tick all applicable boxes.

- [*] I am using Python 3.5.3 or higher (run `python --version` on the command line)
- [*] I have followed the [official guides](https://github.com/Just-Some-Bots/MusicBot/wiki) to install the bot for my system
- [*] I have updated my dependencies to the latest version using the appropriate `update` script

Which version are you using?

- [*] The latest `master` version (release-260419)
- [ ] The latest `review` version

What type of issue are you creating?

- [*] Bug
- [ ] Feature request
- [ ] Question

----

### Description of issue
When I tried to update the file, it says Error: cannot install discord-py[voice] == 1.2.5 and pynacl==1.2.1 because theses package versions have conflicting dependencies, as well as ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
which I have tried and cannot solve the issue of the conflicting dependencies. I have since reverted it into the master version, and below is the whole message.

Checking if we need to update the bot...
Already up-to-date.
Attempting to update dependencies...
Requirement already satisfied: pip in e:\botting\python\lib\site-packages (from -r requirements.txt (line 3)) (20.3.3)
Requirement already satisfied: youtube_dl in c:\users\owner\appdata\roaming\python\python37\site-packages (from -r requirements.txt (line 4)) (2020.12.14)
Requirement already satisfied: colorlog in c:\users\owner\appdata\roaming\python\python37\site-packages (from -r requirements.txt (line 5)) (4.6.2)
Requirement already satisfied: cffi in c:\users\owner\appdata\roaming\python\python37\site-packages (from -r requirements.txt (line 6)) (1.14.4)
Collecting discord.py[voice]==1.2.5
  Using cached discord.py-1.2.5-py3-none-any.whl (655 kB)
Requirement already satisfied: websockets<7.0,>=6.0 in e:\botting\python\lib\site-packages (from discord.py[voice]==1.2.5->-r requirements.txt (line 2)) (6.0)
ERROR: Cannot install discord-py[voice]==1.2.5 and pynacl==1.2.1 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested pynacl==1.2.1
    discord-py[voice] 1.2.5 depends on PyNaCl==1.3.0; extra == "voice"

To fix this you could try to:
1. loosen the range of package versions you've specified
[...]

PR #2180, opened 2021-01-16
Update requirements.txt

Jcodeerd:

- [nope ] I have tested my changes against the `review` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [yes ] I have tested my changes on Python 3.5.3 or higher

----

### Description
removed the versions of the deps to prevent people from manually installing the deps, and the guide / wiki also says that you need to use master, so a suggestion to change this also in master :)

https://github.com/Just-Some-Bots/MusicBot/issues/2176

PS; yeah ik that pr's should target review, but I think this would be handy for master as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
-pynacl==1.2.1
-discord.py[voice]==1.2.5
+pynacl
+discord.py[voice]
 pip
 youtube_dl
 colorlog


srhinos (commented):
We'll merge review into master, this is unnecessary

PR #2183, opened 2021-01-22
Bump min version to 3.8

srhinos:

----

### Description
Bumps min python version, we shouldn't support EOL versions that discord.py won't offer support for as well.


AutumnClove (approved):
I'm pretty sure you hit everything, thank you!
:+1:

PR #2185, opened 2021-01-22
Update windows.md

AutumnClove:
Min version bump

- [x] I have tested my changes on Python 3.8.0 or higher

----

### Description
Changes for min version bump of python.


Jcodeerd (changes requested):
also change `[Python3.7]` to `[python3.9]`?
and `add Python 3.7 to PATH` to `add python 3.9 to PATH`

PR #2187, opened 2021-01-27
Update guides for new min version

### LINKED ISSUE #960: Error installing Pip on Raspbian
I am on a Raspberry Pi 3 B with a fresh image of Raspbian Jessie.  I get an error while trying to install Pip.  I have followed all prior instructions in the Raspbian installation guide and received no errors.  When I run the command `sudo python3.5 get-pip.py` I receive the following error:

> pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.

> Collecting pip
> Could not fetch URL https://pypi.python.org/simple/pip/: There was a problem confirming the ssl certificate: Can't connect to HTTPS URL because the SSL module is not available. - skipping

> Could not find a version that satisfies the requirement pip (from versions: ) No matching distribution found for pip

### LINKED ISSUE #1046: opus_loader not finding libopis on Ubuntu
Having a spot of trouble installing MusicBot on a 12.04 Ubuntu box.  Got all the Python libs installed and everything else, but for some reason it refuses to load the libopis.so.0 module.  The module is clearly installed in the /usr/local/lib folder, so why it can't find it I can't figure.  I also tried putting it directly into the MusicBot folder to no avail.  A screenshots of the filesystem is below.

Oh, and before you come at me asking if I followed the install guide, let me first point out that the PPA for libopis, mc3man, returns a 404 when running sudo apt-get update.  This resulted in me having to compile from source.  Maybe this has something to do with it?  Not sure.  Screenshot of the 404 attached below too.

This is my second attempt at this bot after figuring out that all Raspberry PIs are not created equal.  Lousy gen-1 700MHz CPU....

image

image

### LINKED ISSUE #2343: Update Wiki to include Fedora 38 and Rocky Linux 9 documentation
### What new feature would you like to see?

the wiki is well put together and has a lot of Linux distros on it and mac os and windows etc but the CentOS Documentation is all that exist for RHEL based distro users it would be super helpful to people using Fedora as a Server or people using Rocky and AlmaLinux if they were included in the install guide on the wiki

### Which of these categories fit your request?

Other

### Just checking...

- [X] I have confirmed that my feature has not already been requested.
- [X] I have confirmed that my feature has not already been implemented in the `review` branch.

Jcodeerd:

----


AutumnClove:
Looking good @Jcodeerd I appreciate you doing this!

Jcodeerd:
Thanks!

AutumnClove:
I was thinking of having the intents section added to the configuration page. Not sure if you agree but it seems to be the best place in my opinion to put it. I think all that's left is mac OS?

Jcodeerd:
thats fine, I will add that later these days.
yep, but because I dont have a mac, and cant figure it out with a VM.

Jcodeerd:
@AutumnClove

_docs/installing/aws.md
```diff
@@ -0,0 +1,43 @@
+This walkthrough will take you through the steps required to host the bot externally using a VPS. For this, we recommend the use of Amazon Web Service. You must first have a Personal Account with them before continuing with this walkthrough.
+
+## AWS Installation
+
+Before you can run a program through AWS, you must first create an EC2 Instance. To do this, Navigate to the “Services” tab in the top left and select EC2.
+
+[![](https://aleanaazure.xyz/lib/exe/fetch.php?w=512&h=512&tok=887f2d&media=https%3A%2F%2Fbritishbenji.github.io%2FMusicBot%2Fimages%2FAWS%2FSelect_EC2.PNG)](https://aleanaazure.xyz/lib/exe/fetch.php?tok=a8e437&media=https%3A%2F%2Fbritishbenji.github.io%2FMusicBot%2Fimages%2FAWS%2FSelect_EC2.PNG "https://britishbenji.github.io/MusicBot/images/AWS/Select_EC2.PNG")
```
AleanaA, line 13: Images should be hosted in the `/images` folder. Keeping them hosted elsewhere can lead to dead links, as is the case with this one. This applies to all images listed in this file, as they all seem to be referencing back to the unofficial guide's now defunct URLs referencing back to @BritishBenji 's now dead AWS image links.

_docs/installing/raspbian.md
```diff
@@ -14,17 +14,28 @@ If you're willing to try it, you can run the following commands in order to inst
...
 
+# Installer for python 3.8
+wget https://cdn.discordapp.com/attachments/157598062020132865/804093346007482418/installer.sh # change this later to another link
+# Make it executable
+sudo chmod +x installer.sh
+# Run the script
+./installer.sh
+
+
+
```
AleanaA, line 38: This is a lot of unnecessary white space. We should only have 1-2 spacing lines, not 4.

_docs/installing/raspbian.md
```diff
@@ -14,17 +14,28 @@ If you're willing to try it, you can run the following commands in order to inst
 sudo apt-get update -y
 sudo apt-get upgrade -y
 
+# Installer for python 3.8
```
AleanaA: Although this solution for streamlining Py3.8 installation is good, we can do it one better by piping the script into our shell and exectuting it that way instead. For example:
`curl -s http://example.com/install.sh | sh` would download and execute the `install.sh` script from http://example.com
There's an argument to be made about transparency, but if a user is following a guide like this, odds are they aren't going to read the script anyways.
Regardless, uploading the script to the github page is a better bet than using a discord attachment URL, allowing for the user to review the script on github prior to following the guide, should they wish to.

_docs/installing/raspbian.md
```diff
@@ -14,17 +14,28 @@ If you're willing to try it, you can run the following commands in order to inst
...
+
+
+
 # Install dependencies
-sudo apt install python3-pip
-sudo apt install git
-sudo apt install libopus-dev
-sudo apt install ffmpeg
+sudo apt -y install python3-dev
+sudo apt -y install python3-pip
```
AleanaA: Given we install python 3.8 earlier, we shouldn't need to install python3-dev or python3-pip.
Realistically these are only likely to cause extra conflict, and should pip not be installed with the installer script above, can easily be installed with the get-pip.py script.

_docs/installing/raspbian.md
AleanaA: Along these lines, in line with the argument for transparency, I honestly think the commands to install python should be included in the guide rather than through a script.
Letting the user get a feel for what they're actually doing with their system will help us out a lot in the long run, and if we're going to make automated scripts like this, we should sooner make full bot installation scripts rather than automating the installation of a single dependency.

_docs/installing/scripts/installer_centOS_8.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
AleanaA: Similar to my comments about the Rasbian guide, I'm doubtful scripts like this will do us much good in the long run without just automating the bot's installation process in it's entirety. Attempts have been done to do this and can be found in the review branches main folder, but have not been maintained.

_docs/installing/ubuntu.md
```diff
@@ -70,20 +75,20 @@ sudo add-apt-repository ppa:deadsnakes/ppa -y
...
 cd ~/MusicBot
 
 # Install Python dependencies
-sudo python3.6 -m pip install -U pip
-sudo python3.6 -m pip install -U -r requirements.txt 
+sudo python3.8 -m pip install -U pip
+sudo python3.8 -m pip install -U -r requirements.txt 
 ~~~
 
 ## Ubuntu 14.04
```
AleanaA: We're at a point in time of 14.04's life cycle that we'd be better off removing 14.04 from the guides entirely to clear up space for 22.04 once that releases.

_docs/installing/windows.md
```diff
@@ -3,19 +3,17 @@ title: Windows
...
 MusicBot can be installed on Windows 7, 8, and 10 too, though it requires installing some programs on your computer first.
 
-1. Install [Python 3.7](https://www.python.org/ftp/python/3.7.0/python-3.7.0.exe).
-2. During the setup, tick `Install launcher for all users (recommended)` and `Add Python 3.7 to PATH` when prompted.
+1. Install [Python 3.8](https://www.python.org/ftp/python/3.8.7/python-3.8.7.exe).
+2. During the setup, tick `Install launcher for all users (recommended)` and `Add Python 3.8 to PATH` when prompted.
 3. Install [Git for Windows](http://gitforwindows.org/).
 4. During the setup, tick `Git from the command line and also 3rd-party software`, `Checkout Windows-style, commit Unix-style endings`, and `Use MinTTY (the default terminal MSYS2)`.
 5. Open Git Bash by right-clicking an empty space inside of a folder (e.g My Documents) and clicking `Git Bash here`.
 6. Run `git clone https://github.com/Just-Some-Bots/MusicBot.git MusicBot -b master` in the command window that opens.
```
AleanaA, line 17: It's important to note that the master branch does not support 3.8+ currently.
Each guide that has 3.8 installed rather than 3.7 should be updated to use the review branch instead.

_docs/using/intents.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Intents
+category: Using the bot
+order: 8
+---
+
+#### How to enable Intents?
```
AleanaA, line 7: Although this is a good section, this section doesn't need it's own page, and can be consolidated into the Troubleshooting section of the wiki.

_docs/using/updating.md
```diff
@@ -8,7 +8,7 @@ order: 4
 [![GitHub forks](https://img.shields.io/github/forks/Just-Some-Bots/MusicBot.svg)](https://github.com/Just-Some-Bots/MusicBot/network)
 [![Python version](https://img.shields.io/badge/python-3.5%2C%203.6%2C%203.7-blue.svg)](https://python.org)
 
-> MucicBot version 1.9.8 requires Python 3.5.3 or higher. If you are updating the MusicBot with version below than 1.9.8, reinstall Python with the following version.
+> MucicBot version 02.01.21 requires Python 3.8.0 or higher. If you are updating the MusicBot with version below than 02.01.21, reinstall Python with the following version.
```
AleanaA: Remember to update the links to Python to properly point to 3.8+, rather than pre-3.8 links.

_docs/using/configuration.md
```diff
@@ -8,6 +8,11 @@ Configuring the bot is relatively straight forward. First, open your bot folder
 
 > Editing your configuration file using Notepad or WordPad will result in problems as they save files in strange formats and aren't designed for editing .ini files.
 
+#### Intents
```
AleanaA: This section can be moved to troubleshooting, along with the information about this issue.

_docs/installing/nssm.md
```diff
@@ -0,0 +1,17 @@
+# NSSM
```
AleanaA: Because NSSM and PM2 are more along the lines of launching and keeping the bot running, I feel like these could go more in pages under Using the Bot

_docs/installing/debian.md
```diff
@@ -3,31 +3,35 @@ title: Debian
...
-~~~ bash
+~~~bash
 # Update system repositories
 sudo apt-get update -y
 sudo apt-get upgrade -y
 
 # Install dependencies
 sudo apt-get install git libopus-dev libffi-dev libsodium-dev ffmpeg -y
-sudo apt-get install build-essential libncursesw5-dev libgdbm-dev libc6-dev zlib1g-dev libsqlite3-dev tk-dev libssl-dev openssl -y
+sudo apt-get install build-essential libncursesw5-dev libgdbm-dev libc6-dev zlib1g-dev libsqlite3-dev tk-dev libssl-dev openssl libncurses5-dev libnss3-dev libreadline-dev wget libbz2-dev -y
```
AleanaA: Double check if Curl comes with Debian, and verify if all of these dependencies are necessary. Notably libncurses5-dev and libncursesw5-dev likely share libraries, and we likely don't need both. We shouldn't unnecessarily install new dependencies on a users system and add bloat when it's not needed.

AleanaA (changes requested):
Overall, there's good changes, but there are a number of glaring issues that hold me back from approving this for merge. Especially with Master becoming a stale branch, it's best if we guide people on using Review rather than Master if we want to continue moving forward towards using Python 3.8

_docs/installing/aws.md
Jcodeerd: I think I haven't changed that link or so, just line 29 in this file

_docs/installing/debian.md
Jcodeerd: Curl added incase. Removed libreadline-dev.

Jcodeerd: Wait a sec, why do we need curl actually? This page only uses wget.

_docs/installing/nssm.md
Jcodeerd: Moved.

_docs/installing/raspbian.md
Jcodeerd: Okay, I will "un-script" it and put it here.

_docs/installing/raspbian.md
Jcodeerd: Fixed

_docs/installing/raspbian.md
Jcodeerd: Removed

_docs/installing/scripts/installer_centOS_8.sh
Jcodeerd: Removed the script and put it in the centos page

_docs/installing/ubuntu.md
Jcodeerd: Removed.

_docs/installing/windows.md
Jcodeerd: Done.

_docs/using/configuration.md
Jcodeerd: Done.

_docs/using/intents.md
Jcodeerd: Done.

_docs/using/updating.md
Jcodeerd: Changed to 3.8

_docs/installing/debian.md
AleanaA: > Double check if Curl comes with Debian

I'm more so meaning we shouldn't install wget on a users system if the system already has Curl.

AleanaA (changes requested):
Remember for the several files that lack it, to add the descriptor header as seen in every wiki page. This helps ensure the wiki knows where to place the guide in the list.

AleanaA:
Bumping this as #2283 has been merged, feel free to review the changes made there and update accordingly if still applicable, if not, go ahead and close the PR and I'll handle it when I have time @Jcodeerd

BabyBoySnow:
Is this good to merge? Anything holding it back?

_docs/using/intents.md
```diff
@@ -0,0 +1,22 @@
...
+1. Go to Discord's [applications page](https://discord.com/developers/applications/me) (logging in if prompted).
+2. Select your bot.
+3. Go to the page called "Bot", and scroll until you see "Privileged Gateway Intents"
+4. Now enable "PRESENCE INTENT" and "SERVER MEMBERS INTENT" and you should be good to go!
```
BabyBoySnow: Shouldn't we have them enable message content as well since we use on_message as a command handler?

Jcodeerd: Good one, I'll add it aswell just to be safe

OzzyHelix:
would it be possible to add a guide for doing this in a venv by default Arch Linux wants to use pipx to install stuff in a venv and when you try to install python pip packages to the system you get an error because they want you do run your python apps in a virtual environment

AleanaA:
> would it be possible to add a guide for doing this in a venv by default Arch Linux wants to use pipx to install stuff in a venv and when you try to install python pip packages to the system you get an error because they want you do run your python apps in a virtual environment

That doesn't sound like traditional Arch, that sounds like whatever flavor you're using deviating from the norm. The arch wiki doesn't even have a page for pipx, only a mention on the python page.
I think using venv would be a good idea, but I don't think it's a priority at this time.

Edit: For clarification, traditional arch doesn't really force even pip, normally both python and pip have to be installed by the user just like every other package during the install process

OzzyHelix:
by default in Arch Python base environments are marked as “externally managed” this was introduced with PEP 668
https://peps.python.org/pep-0668

OzzyHelix:
PEP 668 is why I started using venv for the bot

OzzyHelix:
its apparent there to prevent system packages from being broken by pip

PR #2193, opened 2021-02-14
Bumped python version to 3.8 in Dockerfile

phknot:

----

### Description

The python version in the Dockerfile didn't pass the sanity checks. Bumped to 3.8.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.7-alpine
+FROM python:3.8-alpine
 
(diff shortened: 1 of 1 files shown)


Jcodeerd (commented):
Change it to 3.8.7 to be in pair with the new guides?

phknot:
The 3.8 docker tag will always use the latest version within that release. So currently it would indeed use 3.8.7

srhinos (approved):

PR #2236, opened 2021-09-15
Fix plural-s for cmd-clean-reply and cmd-remove-reply

rec0de:

----

### Description

Fixes the logic that determines if a plural "s" is added to item/items or message/messages in `cmd-remove-reply` and `cmd-clean-reply`, previously producing messages like "Cleaned up 0 message".

Somewhat related, I feel like the "item(s)"  in `cmd-remove-reply` (l. 2944, `entry_text`) should be part of the translation string such that only the "s" is added as needed, like for `cmd-clean-reply`. This would allow for better translations that include custom translations for the word "item" but would also require changes to all existing translation files.

If that sounds reasonable to you, I could add the change here and adjust all translation files to produce the same output as before, from where they could be individually updated to provide an actual translation of "item".

Cheers!


AutumnClove:
Looks like a good change to me, all translation files are done by volunteers however and as far as I know we have no sure fire way of communication with them it would be up to users to adjust the translation files as needed, I do think you should add it though, could you also please run black over any changes you make? `python -m pip install black` then `python -m black {file/s}`

rec0de:
Sure can, I'll change this to a draft for now

AutumnClove:
Seeing as you're already making changes, I think perhaps we could additionally add some logic for if no messages are deleted, something like `if len(deleted) < 1: "No messages have been cleared"` of course if you don't feel up to that, that's fine, just a suggestion.

rec0de:
Hmm, I personally feel like that wouldn't be worth the added complexity — especially since you'd still need the "s" check, and it would 'break' existing translations. So I think I'll leave it at that for now, but thanks for the suggestion :)

The last commit I added should not change visible behaviour at all. I've only tested the English version, but the change is basically the same in all files so there shouldn't be any issues.

PR #2243, opened 2021-09-17
Simplify autopause logic

rec0de:

- [x] Auto-formatted changes using `black`

----

### Description

The autopause logic in `on_voice_state_update` is quite complex and distinguishes a lot of different cases, but in the end (as far as I can tell), there are just two possible scenarios: The channel is effectively empty, so we pause, or the channel is not empty, so we resume if we previously auto-paused (or we left the channel, but that's handled separately before).

This patch simplifies that logic and also re-uses existing functions to check for empty channels and perform the auto-pause. It should not change visible behaviour except for some slight changes in info log messages. I have verified that autopause continues to work as expected.

To be honest, I'm not entirely sure when the lines `player.once("play", lambda player, **_: self._autopause(player))` (l.382 & l.736) trigger and what they do, but the code executed for those should not be changed by this patch.

**Update:** Figured it out, should work as expected. Also fixed an edge-case where summoning the bot to a deaf-only channel did not autopause.

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -2801,6 +2801,10 @@ class MusicBot(discord.Client):
 
         log.info("Joining {0.guild.name}/{0.name}".format(author.voice.channel))
 
+        # edge case: we're summoned by a deafened member and should pause on join
+        if self.config.auto_pause:
+            player.once("play", lambda player, **_: self._autopause(player))
+
         return Response(
             self.str.get("cmd-summon-reply", "Connected to `{0.name}`").format(
                 author.voice.channel
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -4159,120 +4159,56 @@ class MusicBot(discord.Client):
         else:
             return
 
-        if (
-            member == self.user and not after.channel
-        ):  # if bot was disconnected from channel
-            await self.disconnect_voice_client(before.channel.guild)
+        if member == self.username:
+            if not after.channel:  # if bot was disconnected from channel
+                await self.disconnect_voice_client(before.channel.guild)
             return
 
         if not self.config.auto_pause:
             return
-
-        autopause_msg = "{state} in {channel.guild.name}/{channel.name} {reason}"
-
-        auto_paused = self.server_specific_data[channel.guild]["auto_paused"]
 
         try:
             player = await self.get_player(channel)
         except exceptions.CommandError:
             return
 
+        autopause_msg = "{state} in {channel.guild.name}/{channel.name} {reason}"
+        auto_paused = self.server_specific_data[channel.guild]["auto_paused"]
+
         def is_active(member):
-            if not member.voice:
-                return False
+            return member.voice and not any(
+                [member.voice.deaf, member.voice.self_deaf, member.bot]
+            )
 
-            if any([member.voice.deaf, member.voice.self_deaf, member.bot]):
-                return False
+        def set_autopause(value):
+            self.server_specific_data[player.voice_client.guild]["auto_paused"] = value
 
-            return True
+        if any(
+            is_active(m) for m in player.voice_client.channel.members
+        ):  # channel is not empty
+            if auto_paused and player.is_paused:
+                log.info(
+                    autopause_msg.format(
+                        state="Unpausing",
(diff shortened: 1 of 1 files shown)


rec0de:
Related question: Is it safe to assume that MusicBot is always a bot? I.e. if `member == guild.me`, is `member.bot` always true?

AutumnClove:
> Related question: Is it safe to assume that MusicBot is always a bot? I.e. if `member == guild.me`, is `member.bot` always true?

Yes, I believe that was added back when users could login with a email and password and just hasn't been touched since, but I might be wrong, it should always be true though.

musicbot/bot.py
```diff
@@ -293,35 +293,35 @@ def _setup_logging(self):
 
     @staticmethod
     def _check_if_empty(
-        vchannel: discord.abc.GuildChannel, *, excluding_me=True, excluding_deaf=False
+        vchannel: discord.abc.GuildChannel, *, excluding_me=True, excluding_deaf=True
```
AutumnClove: we shouldn't exclude deafened members, they can't hear anything anyways

rec0de: the parameter name is a bit confusing, it basically means "exclude deaf members from being counted as active", so the current behaviour is that we don't play for deafened members
perhaps i should change the name

musicbot/bot.py
```diff
@@ -292,36 +292,29 @@ def _setup_logging(self):
...
+        def is_inactive(member):
+            return (
+                member.bot
+                or (not member.voice)
+                or (deaf_as_inactive and (member.voice.deaf or member.voice.self_deaf))
+            )
 
-            if member.bot:
-                return False
+        return all(is_inactive(m) for m in vchannel.members)
```
srhinos, line 304: I love this return statement change but absolutely am not a fan of the refactor of the `_check_if_empty` function as I feel this is much less self documenting as how the function existed before. Would love a middle ground

musicbot/bot.py
```diff
@@ -4159,120 +4150,47 @@ async def on_voice_state_update(self, member, before, after):
         else:
             return
 
-        if (
-            member == self.user and not after.channel
-        ):  # if bot was disconnected from channel
-            await self.disconnect_voice_client(before.channel.guild)
+        if member == self.user:
+            if not after.channel:  # if bot was disconnected from channel
+                await self.disconnect_voice_client(before.channel.guild)
```
srhinos: Needless refactor, should leave it as it was.

musicbot/bot.py
```diff
@@ -4159,120 +4150,47 @@ async def on_voice_state_update(self, member, before, after):
...
-                if auto_paused and player.is_paused:
-                    log.info(
-                        autopause_msg.format(
-                            state="Unpausing",
-                            channel=player.voice_client.channel,
-                            reason="(member undeafen)",
-                        ).strip()
-                    )
+        def set_autopause(value):
+            self.server_specific_data[player.voice_client.guild]["auto_paused"] = value
```
srhinos: This can probably be removed, its only being called once so defeats the line save arguement for nested functions.

musicbot/bot.py
```diff
@@ -2814,6 +2801,10 @@ async def cmd_summon(self, channel, guild, author, voice_channel):
 
         log.info("Joining {0.guild.name}/{0.name}".format(author.voice.channel))
 
+        # edge case: we're summoned by a deafened member and should pause on join
+        if self.config.auto_pause:
+            player.once("play", lambda player, **_: self._autopause(player))
+
```
srhinos, line 2808: Good catch!

srhinos (changes requested):
Did a quick initial pass, once you make a couple changes I'll give it a second more indepth look and test it out!

musicbot/bot.py
rec0de: Ah, you're right - I think I needed the return if `member == self.user` in a previous version but that doesn't seem to be the case anymore after factoring out the empty check

musicbot/bot.py
rec0de: Good catch, that was called twice in a previous version, no need to keep it around

PR #2246, opened 2021-09-20
Enable embeds in now-playing messages, simplify now-playing logic

rec0de:

----

### Description

Use embeds for now-playing messages if enabled in config, rewrite/repurpose dead `update_now_playing_message` function and attempt to simplify some complex logic in `on_player_play` and `update_now_playing_status`.

To streamline things, this uses the same message template for now-playing messages regardless of whether `config.now_playing_mentions` is set or not. The setting now only changes the author's name in the message to a mention.

These changes are live and tested in my customized bot but I have to do some more extensive testing of edge-cases, which is why this is just a draft for now.

*Update:*
Finished testing, also added support for embeds in now-playing messages from the `!np` command and tweaked progress bar appearance.


srhinos:
While I appreciate the PR, I'm not sure this functionality is what I'd want to see in the bot. This removes significant usability and breaks several translations along with taking the bot along a path that was internally discussed and decided against (namely embeds as now playing messages.) 

If you'd like I can leave some comments on your code so you can still get some value out of this exchange tho, review it as if it was going to be merged so maybe you'd still have that oppertunity to iterate on the existing code till its of mergable quality?

rec0de:
Thanks for your feedback!

Would you mind explaining what you mean by removed usability? And is there a reason you chose against using embeds for now-playing messages (even if embeds are enabled in the config)?

Apart from that you don't need to review anything if you don't want it merged.

Issue #2080, opened 2020-05-31
[Feature Request] Support environmental variable for token


khionu:
For the sake of k8s deployment, I'd like to be able to pull my token from the secrets store and insert it as an environmental variable, as a security precaution.

AleanaA:
See #1978

khionu:
I see the PR, and it was rejected for lack of use case and discussion. I have stated my use case. Unless I add a script to my deployment for editing the config file, I have no secure way of inserting my token into the pod on deployment.

srhinos:
I'm gonna tend to agree that this request, while a niche one, is valid.

@khionu, would you mind resubmitting an MR in a similar style and doing a small bit of documenting on it? if not I can work on it in a few days when I have time

khionu:
I guess it'll be a matter of who can find time first, I can't make any guarantees either, with my new job starting tomorrow :D

srhinos:
I'll hit you up on discord if I start working on a fix for this and will edit this comment for other parties interested in this feature.

tripulse:
Khionu tisava '-v

PR #2258, opened 2021-10-05
Add Remote Deployment through Docker

brain-frame:

----

### Description

What is this?
--
This PR adds the functionality to pass envvars into python using `pydantic` which fulfills the request in #2080. Furthermore it adds deployment scripts to push it to a docker-ready remote machine which the localhost has ssh access to.

What changed?
--
- some preferences and options are now configurable via environmental variables prefixed with `MUSICBOT_`
- streamlined the `Dockerfile` so that rebuilds do not take as long (e.g. moving the COPY step to after the installation of dependencies means that that layer doesn't need to be rebuild every time there is a change in the COPY layer)
- Added a deploy script to push data into remote machine

How was this tested?
--
Setting up a musicbot on a remote machine.

- The Discord and Spotify tokens really should be mounted as secrets, but I couldn't find a clean way of mounting them without using K8s or Docker swarm which adds too much complexity on the remote machine side.


AleanaA:
While this PR has good additions such as envvars and fixing up the Dockerfile, remote deployment should be it's own PR
Although it's nice for remote deployement, I'm not sure it has enough use cases to include in the main branch of the bot.
Although the install script in review is in the main branch, realistically these kinds of scripts should be in the gh-pages branch so not to clutter the clone with files that the user won't use after the first install.

BabyBoySnow:
No response from PR'er. After discussion about potentially hosting the scrip(s) on a wiki page, or github pages as mentioned above, it was decided it would be more of a hassle to do for the time being.

PR #2264, opened 2021-10-21
Add docker-compose.yml

RalphORama:

----

### Description

This adds a simple `docker-compose.example.yml` file, which users can copy to `docker-compose.yml` to easily get MusicBot up and running with Docker. The compose file includes a `build` directive to build the local Dockerfile, and a list of `mounts` ported from said Dockerfile.

I named the file `docker-compose.example.yml` and added `docker-compose.yml` to `.gitignore` in case someone wants to make changes to the mount points or add environment variables locally without accidentally adding them to the git repo.

n/a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ config/autoplaylist.txt
 config/autoplaylist_removed.txt
 config/whitelist.txt
 config/blacklist.txt
+
+docker-compose.yml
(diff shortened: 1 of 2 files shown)


AleanaA:
Although this is similar to another recent PR #2258, I prefer this simplistic approach rather than adding things that can be hosted elsewhere.

PR #2283, opened 2021-12-20
Update most guides to Python 3.8+

AleanaA:

----

### Description
Many guides despite the recent merge still point users to the now unsupported Python 3.7
This PR aims to point Linux users to Python 3.8 as this is the current standard for new distributions, and Windows users to Python 3.10
This PR is missing CentOS due to it needing a rewrite, and is also missing MacOS as I lack a mac of any kind to test with and write a guide for it.

### DIFF
--- a/_docs/installing/debian.md
+++ b/_docs/installing/debian.md
@@ -18,7 +18,7 @@ sudo apt-get install git libopus-dev libffi-dev libsodium-dev ffmpeg -y
 sudo apt-get install build-essential libncursesw5-dev libgdbm-dev libc6-dev zlib1g-dev libsqlite3-dev tk-dev libssl-dev openssl -y
 
 # If using Debian Stretch or lower, you need to install Python too using...
-sudo apt-get install python3.5 python3-pip -y
+sudo apt-get install python3.8 python3-pip -y
 
 # Clone the MusicBot to your home directory
 cd ~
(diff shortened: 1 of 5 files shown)


Jcodeerd:
Isnt it the same as https://github.com/Just-Some-Bots/MusicBot/pull/2187 ?

AleanaA:
> Isnt it the same as #2187 ?

Yesn't
#2187 was reviewed prior to activity being done on the main branch, and would pose as a better PR for a proper rewrite of the guides.
This PR aims to be a simple drop in py3.8 replacement, rather than fully fixing the guides for each OS.

Jcodeerd:
Ahh okay. Once this is merged I'll  check up with my pr to make sure it's baseline will be this new pr

_docs/installing/debian.md
```diff
@@ -18,16 +18,16 @@ sudo apt-get install git libopus-dev libffi-dev libsodium-dev ffmpeg -y
 sudo apt-get install build-essential libncursesw5-dev libgdbm-dev libc6-dev zlib1g-dev libsqlite3-dev tk-dev libssl-dev openssl -y
 
 # If using Debian Stretch or lower, you need to install Python too using...
-sudo apt-get install python3.5 python3-pip -y
+sudo apt-get install python3.8 python3-pip -y
```
jayktaylor: I'm not sure that Python 3.8 is readily available in the Debian default repos. Can you confirm? If it isn't, we'll have to build 3.8 from source instead.

_docs/installing/windows.md
```diff
@@ -10,8 +10,8 @@ order: 2
 
 MusicBot can be installed on Windows 7 through 11, though it requires installing some programs on your computer first.
 
-1. Install [Python 3.7](https://www.python.org/ftp/python/3.7.0/python-3.7.0.exe).
-2. During the setup, tick `Install launcher for all users (recommended)` and `Add Python 3.7 to PATH` when prompted.
+1. Install [Python 3.10](https://www.python.org/ftp/python/3.10.1/python-3.10.1-amd64.exe).
```
jayktaylor: Change this to 3.8 so we have the same behaviour on all platforms, even if 3.10 is newer

srhinos (approved):

Issue #2321, opened 2022-12-04
Bot acts up due to TypeError


H3wastooshort:
### What went wrong?

Bot does something like this and goes offline until manually restarted. Happend a few times already

SystemD log after crash:
```
lines 1-21/21 (END)...skipping...
● dc_mus.service - discord music bot
     Loaded: loaded (/etc/systemd/system/dc_mus.service; enabled; vendor preset: enabled)
     Active: active (running) since Sun 2022-12-04 15:11:29 CET; 5h 16min ago
   Main PID: 852080 (bash)
      Tasks: 9 (limit: 1890)
        CPU: 7min 41.182s
     CGroup: /system.slice/dc_mus.service
             ├─852080 bash run.sh
             ├─852096 python run.py
             └─853866 ffmpeg -nostdin -i audio_cache/youtube-JThz8-nYYHs-Nightcore_-_Warriors.webm -f s16le -ar 48000 -ac 2 -loglevel warning -vn pipe:1

Dec 04 18:11:01 lowpower bash[852096]:   File "/mnt/btrfsMirror/discordbots/music/MusicBot/musicbot/bot.py", line 103, in __init__
Dec 04 18:11:01 lowpower bash[852096]:     self.downloader = downloader.Downloader(download_folder="audio_cache")
Dec 04 18:11:01 lowpower bash[852096]:   File "/mnt/btrfsMirror/discordbots/music/MusicBot/musicbot/downloader.py", line 46, in __init__
Dec 04 18:11:01 lowpower bash[852096]:     ytdl_format_options["outtmpl"] = os.path.join(download_folder, otmpl)
Dec 04 18:11:01 lowpower bash[852096]:   File "/usr/lib/python3.9/posixpath.py", line 90, in join
Dec 04 18:11:01 lowpower bash[852096]:     genericpath._check_arg_types('join', a, *p)
Dec 04 18:11:01 lowpower bash[852096]:   File "/usr/lib/python3.9/genericpath.py", line 152, in _check_arg_types
Dec 04 18:11:01 lowpower bash[852096]:     raise TypeError(f'{funcname}() argument must be str, bytes, or '
Dec 04 18:11:01 lowpower bash[852096]: TypeError: join() argument must be str, bytes, or os.PathLike object, not 'dict'
Dec 04 18:11:01 lowpower bash[852096]: [INFO] launcher: All done.
```

### How do we reproduce this?

[...]

PR #2329, opened 2023-08-20
Fix Restart Command & Spotify Credentials

DestinyofYeet:

----

### Description
Fixes the restart command

#2321

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -560,7 +560,7 @@ class MusicBot(discord.Client):
         await self.update_now_playing_status(entry)
         player.skip_state.reset()
 
-        # This is the one event where its ok to serialize autoplaylist entries
+        # This is the one event where it's ok to serialize autoplaylist entries
         await self.serialize_queue(player.voice_client.channel.guild)
 
         if self.config.write_current_song:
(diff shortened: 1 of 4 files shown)


musicbot/bot.py
```diff
@@ -1689,7 +1689,7 @@ async def _cmd_play(
     ):
         player = _player if _player else None
 
-        if permissions.summonplay and not player:
+        if permissions.summonplay:
```
BabyBoySnow: If we don't have `and not player` the constant summon messages when using the play command occur.

DestinyofYeet: Ah ok, not sure how it disapeared but I'll add it back in.

musicbot/bot.py
```diff
@@ -133,6 +133,7 @@ async def _doBotInit(self):
                     aiosession=self.session,
                     loop=self.loop,
                 )
+                await self.spotify.get_token()
```
DestinyofYeet, line 136: This was missing, so it would just not even try to get the token, then complain it didn't have one

BabyBoySnow: Good catch!

BabyBoySnow:
Works fine when restarting with guest mode, however when restarting with ID and secret throw's an error.

PR #2339, opened 2023-09-27
time and date based logging system

OzzyHelix:

- [ ] I have ensured my code is formatted using [Black](https://github.com/psf/black)

----

### Description
I didn't like that the logs wrote over themselves so I made it so log files had dates and times in the file name and when the bot runs it deletes logs older than a week

### DIFF
--- a/musicbot/__init__.py
+++ b/musicbot/__init__.py
@@ -1,6 +1,7 @@
 import sys
 import inspect
 import logging
+from datetime import datetime
 
 from textwrap import dedent
 
@@ -117,7 +118,9 @@ _add_logger_level("VOICEDEBUG", 6)
 log = logging.getLogger(__name__)
 log.setLevel(logging.EVERYTHING)
 
-fhandler = logging.FileHandler(filename="logs/musicbot.log", encoding="utf-8", mode="a")
+# added code for handing logs with dates in the name
+musicbotlog = datetime.now().strftime('logs/musicbot_%Y-%m-%d_%H-%M.log')
+fhandler = logging.FileHandler(filename=musicbotlog, encoding='utf-8', mode='a')
 fhandler.setFormatter(
     logging.Formatter(
         "[{relativeCreated:.16f}] {asctime} - {levelname} - {name} | "
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -9,7 +9,6 @@ import re
 import shlex
 import shutil
 import sys
-import time
 import traceback
 from collections import defaultdict
 from datetime import timedelta
@@ -17,11 +16,10 @@ from functools import wraps
 from io import BytesIO, StringIO
 from textwrap import dedent
 from typing import Optional
-
+from time import time
 import aiohttp
 import colorlog
 import discord
-
 from . import downloader
 from . import exceptions
 from .aliases import Aliases, AliasesDefault
@@ -49,6 +47,13 @@ from .utils import (
 load_opus_lib()
 
 log = logging.getLogger(__name__)
+log.warning("Deleting logs older than 1 week")
+
+# Logs folder path
+path = 'logs/'
+# searches for files that are older than 7 days and deletes them
+result = [os.remove(file) for file in (os.path.join(path, file) for path, _, files in os.walk(path) for file in files) if os.stat(file).st_mtime < time() - 7 * 86400]
+log.info("old logs deleted")
 
 intents = discord.Intents.all()
 intents.typing = False
--- a/run.py
+++ b/run.py
@@ -10,6 +10,7 @@ import logging
 import tempfile
 import traceback
 import subprocess
+from datetime import datetime
 
 from shutil import disk_usage, rmtree
 from base64 import b64decode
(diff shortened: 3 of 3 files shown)


OzzyHelix:
I attempted this before but messed up my fork and had to reset this should be good now

musicbot/__init__.py
```diff
@@ -117,7 +118,9 @@ def _add_logger_level(levelname, level, *, func_name=None):
 log = logging.getLogger(__name__)
 log.setLevel(logging.EVERYTHING)
 
-fhandler = logging.FileHandler(filename="logs/musicbot.log", encoding="utf-8", mode="a")
+# added code for handing logs with dates in the name
+musicbotlog = datetime.now().strftime('logs/musicbot_%Y-%m-%d_%H-%M.log')
```
AleanaA, line 122: We can probably move the date and time to constants.py as STARTTIME, rather than checking the current date and time in every file. Just to ensure any and all uses of the log file use the same file, and don't try to make a new one because they were a second late on a slower device

musicbot/bot.py
```diff
@@ -49,6 +47,13 @@
 load_opus_lib()
 
 log = logging.getLogger(__name__)
+log.warning("Deleting logs older than 1 week")
+
+# Logs folder path
+path = 'logs/'
```
AleanaA, line 53: This seems extraneous since the path is hard coded already

musicbot/bot.py
```diff
@@ -49,6 +47,13 @@
...
+# searches for files that are older than 7 days and deletes them
+result = [os.remove(file) for file in (os.path.join(path, file) for path, _, files in os.walk(path) for file in files) if os.stat(file).st_mtime < time() - 7 * 86400]
```
AleanaA, line 55: This should be a config option, rather than default, permanent behavior. Some users may, even if unnecessary, find it important to keeps older logs to find a song from a user again, or if they want to keep logs as a result of tracking a rare issue

run.py
```diff
@@ -145,16 +146,10 @@ def get_requirements(cls, file="requirements.txt"):
...
-    if os.path.isfile("logs/musicbot.log"):
-        log.info("Moving old musicbot log")
-        try:
-            if os.path.isfile("logs/musicbot.log.last"):
-                os.unlink("logs/musicbot.log.last")
-            os.rename("logs/musicbot.log", "logs/musicbot.log.last")
-        except:
-            pass
+    # added code for handing logs with dates in the name
+    musicbotlog = datetime.now().strftime('logs/musicbot_%Y-%m-%d_%H-%M.log')
```
AleanaA, line 150: Same as above, it doesn't make sense to check the current time in different places when it's supposed to be the startup time.

AleanaA (changes requested):
Just a couple notes before considering merging, the idea overall is better though, and should make logs significantly easier to track

BabyBoySnow:
https://docs.python.org/3/howto/logging-cookbook.html#using-file-rotation

In addition to what AleanaA has said, using a rotating file handler would be preferable.

OzzyHelix:
I will retract my pull request then I'm not very experienced with python and this was the best I could do with my current skill level sorry I tried to help but I'm just not good enough to improve this

OzzyHelix:
I guess my goal with this was to share my code even tho its not that great

OzzyHelix:
would it be a problem if I made a feature request for this in issues?

OzzyHelix:
if I made a feature request for this. I believe that you would have a better implementation than what I did

OzzyHelix:
assuming that feature request goes anywhere

OzzyHelix:
I figured I should ask if its ok to make an issue feature request regarding this I believe that the logging system could be better and I attempted to fix it but I just lack the skill to do it fully so I thought making a issue feature request would be reasonable I just wanted to run that idea by you guys to see if that would be ok

BabyBoySnow:
You can open a fr, however I'll be honest the likelihood of me getting to this one any time soon is low. I was rather hoping you'd implement the requested changes

OzzyHelix:
I thought what I did was clever but its more that I just don't know what I'm doing I'm sorry for wasting you time

BabyBoySnow:
It's never a waste of time. You're taking the time to learn, when I first started using MusicBot I didn't even know how to do anything in Python. There's still lots of things I don't know. If you are unfamiliar with the requested changes, take some time to read up and improve.

OzzyHelix:
the only reason I have been able to do stuff with your bot is because its laid out in a way that I can understand I am not familiar with python but I can edit things within your bot to suit specialized use cases I have made the mistake before of being a hostile fork but I do not wish to repeat that my fork only exist now to stay up to date with the dev branch

BabyBoySnow:
I apologize if I came off as hostile, that was not my intention I was trying to be positive and encourage learning.

OzzyHelix:
its ok I should leave development to people who know what they're doing

BabyBoySnow:
Please do not become discouraged from our request for change. I highly encourage PR's and people to contribute. Our request are mostly due to the fact that we may not always have time to get around to stuff. Overall this is a great start! It just needs some adaptation for a wider audience. For example if someone was using this with a slow system they might get two different times from one instance.

PR #2351, opened 2023-10-26
Move message fix

DestinyofYeet:
en.json: fixed spelling mistakes

----

### Description
Fixes the formatting of the move message. `.format()` has to be called on the result of `self.str.get()`, as that returns the string from the en.json file and we need to format that.


BabyBoySnow:
thanks

PR #2355, opened 2023-11-03
Fix Docker build (Include git in Dockerfile)

phknot:

Changed are tested in the dev branch, might be useful to merge this commit to review/master as well.

- [x ] I have tested my changes on Python 3.8 or higher
- [x ] I have ensured my code is formatted using [Black](https://github.com/psf/black)

----

### Description

git is needed to install the git+https://github.com/Rapptz/discord.py pip dependency.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,7 +18,8 @@ RUN apk update && apk add --no-cache \
   opus-dev \
   libffi \
   libsodium \
-  gcc
+  gcc \
+  git
 
 # Install pip dependencies
 RUN pip3 install --no-cache-dir -r requirements.txt


BabyBoySnow:
Is there any reason for this?

BabyBoySnow:
I think users can do this themselves, since it's only that way until a new pypi is released.

phknot:
> I think users can do this themselves, since it's only that way until a new pypi is released.

That would assume they know how Docker works. I also find it a bit weird to leave one of the possible deployment methods just in a broken state in the repository, when it's an easy fix.

BabyBoySnow:
> > I think users can do this themselves, since it's only that way until a new pypi is released.
> 
> That would assume they know how Docker works. I also find it a bit weird to leave one of the possible deployment methods just in a broken state in the repository, when it's an easy fix.

Yes I see where you're coming from, my thing is that it would probably just get changed back when new pypi is released, however on second thought I might just keep it this way to have this branch always up to date with the latest discord.py, so we can have stuff fixed before being released to master. I'm going to ask the other members what they think before merging however. Thanks for your input it's appreciated. If you have anymore pr's please submit them and I'll look over them :)

PR #2361, opened 2023-11-27
Round robin queue

DestinyofYeet:

----

### Description

Added a round robin queue


config/example_options.ini
```diff
@@ -169,6 +169,9 @@ LeaveInactiveVCTimeOut = 300
 # Sets if if the bot should leave once all songs have finished playing.
 LeaveAfterSong = no
 
+# Round robin queue
+RoundRobinQueue = yes
```
BabyBoySnow: Should be set to no by default to match the default settings in config.py 
Would prefer more descriptive comment about the functionality.

Issue #1817, opened 2019-01-12
Keep Duration/Total file size limit for the downloaded audio files


felix920506:
Please tick all applicable boxes.

- [x] I am using Python 3.5 or higher (run `python --version` on the command line)
- [x] I have followed the [official guides](https://github.com/Just-Some-Bots/MusicBot/wiki) to install the bot for my system
- [x] I have updated my dependencies to the latest version using the appropriate `update` script

Which version are you using?

- [x] The latest `master` version (1.9.8)
- [ ] The latest `review` version

What type of issue are you creating?

- [ ] Bug
- [x] Feature request
- [ ] Question

----

### Description of issue
Saving the downloaded audio files takes up a lot of space but not saving it causes delays when playing songs. Can you add the option to delete the audio files after it reaches a certain duration without being played or when the folder reaches a set limit in size

AleanaA:
Duplicate of #1584

Issue #394, opened 2016-06-29
[Suggestion] Implement audio cache auto cleanup feature


strongestFairy:
Currently I'm hosting MisicBot at VPS with tiny 5GB drive. It was completely filled out with audio_cache in one month and bot became unusable (impossible to add new music requests).

Suggestion: implement auto cleanup feature for the auto_cleanup folder. Of course it can be done by external script but embedded feature looks better here. We need to add some _cache_max_size_ parameter and implement Bot behavior when limit is overused. For example delete some fixed percent or number of most unused/most old files.

arunesh90:
https://github.com/SexualRhinoceros/MusicBot/blob/master/config/example_options.ini#L72

strongestFairy:
> https://github.com/SexualRhinoceros/MusicBot/blob/master/config/example_options.ini#L72

This is not what I mean actually. This option completely disables caching. The good solution is to cache often used files and delete unused if they overcome the storage size limit.

zstinnett:
Is there any update on when this might be implemented? I would like to be able to leave alone, self regulating.

strongestFairy:
Well... I did my solution with bash script.
Just set proper value of TARGET_DIR and desirable MAX_DIR_SIZE in bytes.
Script should be launched via cron daemon like

```
# m h  dom mon dow   command
0 5 * * * /opt/cleanup/clean_audio_cache.sh 2>&1 >/dev/null
```

```
#!/bin/bash

TARGET_DIR='/opt/MusicBot/audio_cache'
MAX_DIR_SIZE=1500000

if [ ! -d $TARGET_DIR ]; then echo "$TARGET_DIR does not exists"; exit; fi
cd $TARGET_DIR
DIR_SIZE=`du -s . | cut -f 1`

if [ "$DIR_SIZE" -gt "$MAX_DIR_SIZE" ]
    then
    echo "$DIR_SIZE > $MAX_DIR_SIZE"
    echo "Delete extra files"
    ls -ltu | tail -n 100 | tr -s ' ' | cut -d ' ' -f 9 | xargs rm
    echo "Done!"
    else
    echo "$DIR_SIZE < $MAX_DIR_SIZE"
    echo "Nothing to delete"
fi
```

zstinnett:
I was able to add it to my crontab it have it run every Thursday! Thank you!

PR #2362, opened 2023-11-28
More audio cache options.

itsTheFae:
- [x] I have ensured my code is formatted using [Black](https://github.com/psf/black) & Flake8

----

### Description

This PR primarily improves automatic cache management when `SaveVideos` is enabled.  It adds two new options, `StorageLimitBytes` and `StorageLimitDays` to control size and age of cache data.  Cache files are sorted by their access-time in *nix/MacOS and by creation-time in Windows, to remove least-recent-used or oldest songs before others.  
Caches are purged on startup as before.  However, if `StorageLimitBytes` is set and the cache exceeds this value it will be purged when a song is played to maintain the set limit.  
Additionally a new owner-only command `!cache` was added.  Using `!cache` or `!cache info` reports cache size and settings information.  Using `!cache clear` will manually clear the cache.  
When `SaveVideos` is disabled, Windows should now actually delete the downloaded file after playback. The cache directory is still deleted and recreated at startup as with previous versions. 

After reading PR #1884 I am tempted to add a Min and Max replacement for `StorageLimitBytes` option.  To reduce filesystem calls as much as possible by purging down to Min bytes rather than down to `StorageLimitBytes` and prevent every new song played from doing cache checks when cache fills.   I haven't tested this on a giant cache, so I don't know if it would be needed or not...

Feedback, edits, or cherry picking all welcome.  Hope it's useful.

#394 
#1584 
#1817


musicbot/bot.py
```diff
@@ -3624,6 +3691,78 @@ async def cmd_option(self, player, option, value):
...
+        """
+        Usage:
+            {command_prefix}cache
+
+        Display cache storage info or clear cache files.
+        Valid options are:  info, clear
+        """
+        opt = opt.lower()
+        valid_opt = ["info", "clear"]
+        if opt not in valid_opt:
```
DestinyofYeet: From Just Snow: I'd prefer `if opt not in ["info, "clear"]` then throw an invalid usage error

DestinyofYeet: What platform did you develop and test this code on btw?

itsTheFae: > What platform did you develop and test this code on btw?

This was tested on Ubuntu 22.04 server.  I have not tested on windows and I could not confirm via the python docs and a little search if `getatime()` is a problem for windows or NTFS.  It _should_ work, but I'd have to load a VM to test it.

BabyBoySnow:
Could you join the support server it's a little easier to communicate there, I'm on windows and can test here in a bit.

config/example_options.ini
```diff
@@ -70,6 +70,17 @@ SkipRatio = 0.5
 # If "yes", videos will not be re-downloaded if queued again, at the cost of hard-drive space.
 SaveVideos = yes
 
+# Set a time limit for stored files. Set to 0 to disable.
+# Files which aren't used for this period of time will be removed.
+# Only applies when SaveVideos option is enabled.
+StorageLimitDays = 182
```
BabyBoySnow: Let's leave this off by default

itsTheFae:
> Could you join the support server it's a little easier to communicate there, I'm on windows and can test here in a bit.

I would join but I like to keep my phone number private.  That appears to be a problem in this instance.

BabyBoySnow:
> > Could you join the support server it's a little easier to communicate there, I'm on windows and can test here in a bit.
> 
> I would join but I like to keep my phone number private. That appears to be a problem in this instance.

I understand. That being said, what was the testing parameters you used? I'll double check on windows to make sure everything is working as intended, as for mac, we'll just have to wait and see. 
You can send me a friend request on discord as well if you'd like. @therealjustsnow

DestinyofYeet:
It's not a requirement, we can also just communicate here. We currently still have an issue with cached files not being deleted correctly, that's why I was asking the platform.

itsTheFae:
@BabyBoySnow  Not sure I understand exactly what parameters you are asking for.  The two new configs were tested at a few different values.  I set `StorageLimitDays` to values inclusively between 6 to 3 to verify files older than that got removed.  The `StorageLimitBytes` option I have tested at a lower value of `20971520` or 20MB for most of the testing.  


@DestinyofYeet  I have not yet experienced any issues removing the files individually at least. Granted the bot isn't serving multiple guilds while I'm testing so I could be missing out on some race conditions.  Two guilds play the same song at the same time, or one guild plays a song that another guild just removed from cache or the like...

DestinyofYeet:
We encountered an issue where the bot wasn't able to delete the file, since it was still "In use" according to windows at least.

BabyBoySnow:
The issue is a windows one apparently as it doesn't seem to happen on linux. I believe the root cause is the subproccess taking a bit to release, giving an "file is in use" error when attempting to delete. This was mostly confirmed by comparing the time the file was beint attempted to be deleted to the time that ffmpeg was successfully terminated. Also that's the parameters I wanted to know, it's best if conditions are as closely replicated as possible between test ya know? Ensures no funky business.

itsTheFae:
Understood.  Linux kind of queues the file delete, so if a process is using the data it isn't deleted until that process closes the file descriptor.  Windows on the other hand does have file lock quirks, and Virus protection complication on top of that. 

If the code is problematic in windows, we can probably get away with just catching and ignoring errors.  At a later point when cache is cleared again the file may not be locked so it will be deleted then.  Might be better/faster than retrying right away anyhow.

itsTheFae:
Dusted off an old Windows 10 VM, so I'll test this out and post back with needed changes.

itsTheFae:
So this should work.  Had to compromise on windows with using ctime rather than atime.  It appears that atime is updated for every file that ffmpeg has played, each time it plays a new file.  Not sure why or how, but it does.  
I've tested this with `SaveVideos` disabled as well, and opted to prevent whole-directory deletion post-play, it only happens on startup now.  

I need to dig around more in player.py, as far as I understand it player.py is the ffmpeg process, so doing any file clean up there probably isn't going to work.   We need a way to make sure ffmpeg is closed before doing any file operations.  ~Not sure if just waiting until `on_player_stop` would work or not.~  Almost sure it wont.

BabyBoySnow:
Lmk when I should merge this, seems you're still experimenting.

itsTheFae:
@BabyBoySnow  I believe I have fixed file deletion on windows while `SaveVideos` is disabled.  It should be ready for merge now.

BabyBoySnow:
> @BabyBoySnow I believe I have fixed file deletion on windows while `SaveVideos` is disabled. It should be ready for merge now.

Gotcha I'll go ahead and merge it on a whim then, PC is blue screening so can't really test haha. If anything I'll boost activity in the support server if it breaks lmao

Issue #2288, opened 2022-01-18
Set prefix per server


andreyrd:
### What new feature would you like to see?

I would like the ability to set custom prefixes for each server the bot is in. The the UI for this can be something like `{current prefix}setprefix {new prefix}`.

### Which of these categories fit your request?

New command

### Just checking...

- [X] I have confirmed that my feature has not already been requested.
- [X] I have confirmed that my feature has not already been implemented in the `review` branch.

PR #2367, opened 2023-12-12
Prefix per server

itsTheFae:
- [x] I have tested my changes on Python 3.10.12

----

### Description

This PR adds support for setting a command prefix per guild/server.  It adds one new option `EnableOptionsPerGuild` and a new command `setprefix` to manage the feature.  
This feature allows just about anything without spaces to be set as a command prefix, including Emoji.  Only one verb cannot be set, as `setprefix clear` is used to clear any currently set prefix.
The default prefix in `options.ini` is used when no server-specific prefix is set.  If per-guild options are not enabled the `setprefix` command will raise an error.  

Additionally, this feature will remember a limited number of prefix changes so the `clean` command can take care of messages prior to a prefix change.   This history of prefixes does not persist through restarts.

#2288


config/example_options.ini
```diff
@@ -166,6 +166,9 @@ SearchList = no
 # Set the amount of search results returned by default.
 DefaultSearchResults = 3
 
+# Allow per-guild command prefix override via 'setprefix' command.
+EnableOptionsPerGuild = no
```
DestinyofYeet: Personally I'd think something like `EnablePrefixPerGuild` would be better.

DestinyofYeet (commented):
Looks good

config/example_options.ini
itsTheFae: I agree.  I started this intending to add more options but ended up just doing command prefix to keep it focused.

itsTheFae:
This should be fully ready to go now. :+1:

Issue #2368, opened 2023-12-13
New caching code doesn't work on Python 3.8


phknot:
### What went wrong?

The new code introduced with https://github.com/Just-Some-Bots/MusicBot/pull/2365 uses arrow notation in function definitions, which is only supported since Python 3.9.

This project advertises a minimum version of 3.8 to function. I suggest either modifying the code to work on Python 3.8, or updating the whole project to 3.9

### How do we reproduce this?

Try running the bot with Python 3.8, for example with the Docker image. It will give an error.

### Which version of the bot are you using?

None

### Which operating system are you using?

None

### Just checking...

- [X] I have confirmed that my issue has not already been reported.
- [X] I am using Python 3.8 or higher.
- [X] I have updated my Python dependencies.

itsTheFae:
So I checked the docs and they seem to indicate that type hinting with arrows was added in 3.5:  https://docs.python.org/3.8/library/typing.html  

Do you mind documenting what OS you're on, and posting a copy of the error log for us to review?  Thanks.

phknot:
> So I checked the docs and they seem to indicate that type hinting with arrows was added in 3.5: https://docs.python.org/3.8/library/typing.html

Type hinting on collections was added in 3.9 https://docs.python.org/release/3.9.0/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

> Do you mind documenting what OS you're on, and posting a copy of the error log for us to review? Thanks.

I run the bot using the provided docker setup. When trying to start you'll get the following error:

```
musicbot-musicbot-1  | [ERROR] launcher: Error starting bot
musicbot-musicbot-1  | Traceback (most recent call last):
musicbot-musicbot-1  |   File "run.py", line 409, in main
musicbot-musicbot-1  |     from musicbot import MusicBot
musicbot-musicbot-1  |   File "/musicbot/musicbot/__init__.py", line 4, in <module>
musicbot-musicbot-1  |     from .bot import MusicBot
musicbot-musicbot-1  |   File "/musicbot/musicbot/bot.py", line 33, in <module>
musicbot-musicbot-1  |     from .filecache import AudioFileCache
musicbot-musicbot-1  |   File "/musicbot/musicbot/filecache.py", line 19, in <module>
musicbot-musicbot-1  |     class AudioFileCache:
musicbot-musicbot-1  |   File "/musicbot/musicbot/filecache.py", line 40, in AudioFileCache
[...]

PR #2369, opened 2023-12-13
Update Python versions

phknot:

- [x] I have tested my changes on Python 3.9 or higher

----

### Description

The minimum Python version is bumped to 3.9 to fix compatibility with new code.

Where possible, we now try to use the latest Python version (3.12).

https://github.com/Just-Some-Bots/MusicBot/issues/2368

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -50,7 +50,7 @@ body:
       options:
         - label: I have confirmed that my issue has not already been reported.
           required: true
-        - label: I am using Python 3.8 or higher.
+        - label: I am using Python 3.9 or higher.
           required: true
         - label: I have updated my Python dependencies.
           required: true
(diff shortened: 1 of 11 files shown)


itsTheFae:
I would suggest keeping 3.8 in this PR.  I've fixed my typing blunder with #2370  which will let it work again.  
Or at least its an option. :P

phknot:
> I would suggest keeping 3.8 in this PR. I've fixed my typing blunder with #2370 which will let it work again. Or at least its an option. :P

Yeah I guess a maintainer can make a decision on how to proceed 👍

BabyBoySnow:
By the way, as far as I know aiohttp a required dependency is still failing on build wheels on 3.12, I'm hoping they fix that soon. 
For the time being, I will continue backwards compatibility, until told otherwise from the code owners.

phknot:
> By the way, as far as I know aiohttp a required dependency is still failing on build wheels on 3.12, I'm hoping they fix that soon. For the time being, I will continue backwards compatibility, until told otherwise from the code owners.

I tested all the versions by bumping the version in the Docker file and it ran just fine for all.python versions from 3.9 to 3.12 so I think that problem is fixed

BabyBoySnow:
> I tested all the versions by bumping the version in the Docker file and it ran just fine for all.python versions from 3.9 to 3.12 so I think that problem is fixed

Using windows 10 22h2, I still fail the build wheel on 3.12.1. The "solution" is to install Microsoft C++ Build Tools. I don't like this as a solution though since it's a 4-6gb download, which is a lot for users who might not have a ton of space on a vps. It builds fine on 3.8-3.11 without the additional tools. So I don't think it's anything on our end in the project.

PR #2378, opened 2023-12-20
Stop spotify playlists fetch on clear command

### LINKED ISSUE #2375: Spotify playlists take time time to fetch, rendering !clear kinda useless.
### What went wrong?

Maybe this should qualify as a feature request, I'm not sure. Tested on the dev branch (latest commit).

Spotify playlists take a while to be cached by the bot; it takes a couple seconds for each individual song to be put into the queue. 

The issue arises when someone with permissions wants to clear the queue of this playlist. It effectively becomes impossible without restarting the bot or without waiting several minutes for the songs to be cached. This is because e.g. if 5 songs have been put into the queue, when someone uses `!clear` the bot will clear the queue BUT it will also keep adding new songs to the queue (songs that were unfetched at the time). Only way to stop the bot is to restart it.

I imagine the ideal solution would be to make `!clear` cancel any pending attempts to fetch songs from playlists.

### How do we reproduce this?

1) On a bot instance with Spotify credentials enabled, use `!play [spotify playlist URL]` in which the playlist has multiple songs (more than 5 is fine)

2) (Optional) Once the first song starts playing, observe with `!queue` that only one or two songs are queued at this moment. Over time more songs will be added; next steps must be taken before the whole playlist is queued (which takes a while anyway)

3) Use `!clear` in an attempt to clear the queue

4) Observe that while the queue at the time was cleared, this doesn't stop the bot from fetching more songs from the playlist that weren't queued at the time

### Which version of the bot are you using?

I am using the "dev" branch

### Which operating system are you using?

Ubuntu 18.04 or higher

### Just checking...

- [X] I have confirmed that my issue has not already been reported.
- [X] I am using Python 3.8 or higher.
- [X] I have updated my Python dependencies.

itsTheFae:
- [x] I have tested my changes on Python 3.8.10 on windows and 3.10.12 on linux

----

### Description

Adds yet another server-specific data entry to stop playlist loops from queuing songs after a `clear` command has been issued.   This PR has only focused on the spotify lists, since those seem the worst offender.   
Please inform if other playlist logic should be checked! 

In short, when command `clear` is issued the halt flag is set, and it will be cleared when the next call to play a song is made. 

closes #2375

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -124,6 +124,7 @@ class MusicBot(discord.Client):
             "last_np_msg": None,
             "availability_paused": False,
             "auto_paused": False,
+            "halt_playlist_unpack": False,
             "inactive_player_timer": (
                 asyncio.Event(),
                 False,  # event state tracking.
@@ -2321,8 +2322,19 @@ class MusicBot(discord.Client):
                                 ).format(res["name"], song_url),
                             )
                             for i in res["tracks"]["items"]:
+                                if self.server_specific_data[channel.guild.id][
+                                    "halt_playlist_unpack"
+                                ]:
+                                    log.debug(
+                                        "Halting spotify album queuing due to clear command."
+                                    )
+                                    break
                                 song_url = i["name"] + " " + i["artists"][0]["name"]
-                                log.debug("Processing {0}".format(song_url))
+                                log.debug(
+                                    "Processing spotify album track:  {0}".format(
+                                        song_url
+                                    )
+                                )
                                 await self.cmd_play(
                                     message,
                                     player,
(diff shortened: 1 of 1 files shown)


BabyBoySnow:
I don't see any reason to just make this for spotify, might as well make it work for any event where clear is issued and song(s) are still being downloaded

itsTheFae:
After reading and poking through the code I have concluded that this feature cannot currently be extended to the other playlist types that are actually handled by ytdl.  At least, not without restructuring how we interface with ytdl.  

While all playlists take time to process and download, those handled by ytdl appear to queue all extracted URLs at once when the extraction process is finished.   Only the spotify playlist implementation suffers from the continuous, sequential queuing issue, which this patch improves but does not 100% solve.  

In my testing I have found that some of the time, issuing a `clear` command would not halt the playlist unpack.  This is a result of the flag to halt playlist unpack being unset on a subsequent play call, and the current playlist implementation directly calls play to enqueue the track search terms.   A race condition that cannot be avoided without restructuring how we deal with spotify.   
Issuing multiple `clear` commands typically overcomes this, and I think this is acceptable for the time being considering that without this patch, one must issue clear an equal number of times to the number of playlist tracks.  So a potential reduction from dozens or hundreds to only 2 or 3 commands. 

With all this said, I think this PR is as good as it is going to get for now and is ready for merge.

DestinyofYeet (approved):
Looks good

PR #2379, opened 2023-12-22
Update flake8 config and run Flake8 over the whole project.

itsTheFae:

----

### Description

This PR updates some flake8 configs and brings the source into compliance with flake8.  Primarily addresses the various bare `except` statements that could be catching system interrupt signals.  
Some effort was made to use more specific Exceptions where they could be used.  However, some except statements may still be too broad, catching and eating `asyncio.exceptions.CancelledError` where they maybe should not be ignored.

Also changes skip command permission checks for looped songs, since clean up made it throw command permissions error.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,5 +1,6 @@
 [flake8]
-ignore = E203, E266, E501, E731, W503, F403, F401
+ignore = E203, E266, E501, W503, C901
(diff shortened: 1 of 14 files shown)


DestinyofYeet (approved):
Looks good to me

PR #2398, opened 2024-05-01
Replaced string.format () and % with f-strings. 

chriszaro:
Pylint reports "[consider-using-f-string". f-string is faster, more readable and futureproof.
Pylint now reports about 300 less conventions. There are more to be refactored.

image

- [ ] I have ensured my code is formatted using [Black](https://github.com/psf/black)

----

### DIFF
--- a/bootstrap.py
+++ b/bootstrap.py
@@ -73,16 +73,16 @@ MINIMUM_PY_VERSION = (3, 8)
 TARGET_PY_VERSION = "3.8.7"
 
 if SYS_PLATFORM not in PLATFORMS:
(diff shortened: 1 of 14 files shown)


BabyBoySnow:
Hello, actually the dev branch is already py linted. We're just holding off on putting it on master for the reason that sev currently uses the most up-to-date discord.py, which hasn't been released on pypi yet so we have changes that won't work. 
Additionally, please review our pull request guidelines, you've targeted the wrong branch, pull request should target the dev branch. 
Additionally you haven't said if you've tested these changes.

PR #2401, opened 2024-05-14
Installer updates

itsTheFae:
- [x] I have tested my changes on Python 3.10 or higher

----

### Description

These enable the install scripts to detect when they are run from a repo clone and continue installation without prompting for a new branch / clone to be made.
Some minor changes to output in the windows ps1 script to make it a tad less verbose and clear to see what is going on. 
This also fixes an edge case with startup when discord.py is missing but an extension package is installed.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -100,9 +100,10 @@ else
 }
 ""
 
-# TODO: fix this, it fails due to some missing addition to powershell environment.
-# idk if it is even needed, but was here before so whatever I guess...
-Invoke-Expression "refreshenv" 2>$null
+# NOTE: if we need to refresh the environment vars (Path, etc.) after installing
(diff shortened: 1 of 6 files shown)


install.ps1
```diff
@@ -53,82 +53,102 @@ $DEFAULT_URL_BASE = "https://discordapp.com/api"
...
 {
-    "FFmpeg already installed"
+    "FFmpeg already installed."
 }
+""
 
 # TODO: fix this, it fails due to some missing addition to powershell environment.
-Invoke-Expression "refreshenv"
+# idk if it is even needed, but was here before so whatever I guess...
+Invoke-Expression "refreshenv" 2>$null
```
BabyBoySnow: Yeet this

BabyBoySnow (changes requested):

PR #2405, opened 2024-05-17
Add the follow command

itsTheFae:
- [x] I have tested my changes on Python 3.10 or higher

----

### Description

New command `follow` allows MusicBot to follow a user between channels in a guild.

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -504,6 +504,12 @@ class MusicBot(discord.Client):
                 channel_map[guild] = guild.me.voice.channel
                 resuming = True
 
+            # Check for follow-user mode on resume.
+            follow_user = self.server_data[guild.id].follow_user
+            if from_resume and follow_user:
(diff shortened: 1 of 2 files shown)


musicbot/bot.py
```diff
@@ -4673,6 +4679,56 @@ async def cmd_summon(
...
+            {command_prefix}follow
+
+        MusicBot will automatically follow a user when they change channels.
+        """
+        # If MusicBot is already following a user, either change user or un-follow.
+        followed_user = self.server_data[guild.id].follow_user
+        if followed_user is not None:
+            # Un-follow current user.
+            if followed_user.id == author.id:
+                # TODO:  maybe check the current channel for users and decide if
```
BabyBoySnow, line 4699: isn't this already added below lol?

itsTheFae: This would need to be tested.  It might need some specific logic in the command if the logic in voice-state-update does not cut it.

PR #2408, opened 2024-05-18
Provide option for auto-unpause, default off.

itsTheFae:
- [ ] I have tested my changes against the `dev` branch (the latest developmental version), and this pull request is targeting that branch as a base

----

### Description

Adds a config option `UnpausePlayerOnPlay` to control automatic playback resume when play commands are used.
Default is off to return MusicBot to its previous behavior.

This has not been tested, more changes might be needed.


BabyBoySnow:
Needs to be added to example_options.ini

itsTheFae:
I seem to always forget that the file exists and needs updates. lol

Issue #2413, opened 2024-06-12
aliases with spaces don't work


Jonas1999:
### What went wrong?

I'm trying to get `"restart upgrade": ["upgrade", "update", "test"]` in aliases.json to work but it doesn't work.
My other aliases work except that one, the entire aliases.json contains the following:
`{
    "play": ["p"],
    "skip": ["s"],
    "queue": ["q"],
    "latency": ["ping"],
    "restart": ["reboot", "reset"],
    "restart upgrade": ["upgrade", "update", "test"],
    "checkupdates": ["checkupdate"],
    "summon": ["join", "connect", "come"],
    "disconnect": ["leave", "stop"]
}`

### How do we reproduce this?

Have an aliases.json contain what I mentioned above and run the bot, the aliases won't be acknowledged by the bot.
The bot is running on an Raspberry Pi 4 with Debian 12/ Raspberry Pi OS 64 Bit.

### Which version of the bot are you using?

I am using the "review" branch

### Which operating system are you using?

Ubuntu 18.04 or higher

### Just checking...

- [X] I have confirmed that my issue has not already been reported.
- [X] I am using Python 3.8 or higher.
- [X] I have updated my Python dependencies.

itsTheFae:
Currently, aliases are quite limited and will only function as replacement names for command names.  

I did poke around with this while working on the last few rounds of updates.  I think the feature could definitely use a handful of changes.   For example, being able to manage the aliases via command would be nice to have, or minimally some indicator that the aliases even exist via like help command responses.  

At least I can say it was on my radar.  Was going to take it on before doing the UI stuff I had planned. But it could be a while before I get the time to work out proper updates for it.

BabyBoySnow:
I'm closing this as complete due to the latest pr merged into dev, if it still doesn't meet you're requirements feel free to open a new issue

PR #2416, opened 2024-07-23
Bug-fixes and more options.

itsTheFae:
- [x] I have tested my changes on Python 3.8 or higher (3.8.10 on Win 10 & 3.10.12 on PopOS 22.04)

----

### Description

This PR contains a few patches for stability, some corrections to features, and also adds new config options to tweak some of ytdlp's network related settings.   
Some of these patches will need some extra testing on windows or with python 3.8 to be sure they work.  As always, feel free to deny this PR and cherry-pick the parts that are most useful.  

A summary list of changes:  
- Add support for `cookies.txt` by adding the file `MusicBot/data/cookies.txt` to enable cookies.  Must not be empty!
- Add option to allow changing yt-dlp UA strings from dynamic to static / custom UA.
- Add option to enable HTTP/HTTPS proxy for yt-dlp and media checking only.  (APIs and ffmpeg will not use this proxy.)
- Add option to count paused players, off by default to restore historic behavior of status message.
- Add support for complex aliases, with arguments.  
- Re-enables pre-download for next track in the queue. (still no support for auto playlist tracks and the new-song edge case.)
- Better handling of player inactivity checks to prevent premature disconnects.
- Better handling of start-up failures to trigger installing dependencies.
- Adds a check for ffmpeg being executable to start-up.
- Windows will now prefer system-installed ffmpeg over using bundled `bin/` exes.
- Improved debug command flexibility.
- Added dev-only command to generate markdown from config and permissions code.  A step to generating documentations from the source.

Alias support was expanded on, potentially closing issue #2413

### DIFF
--- a/config/example_aliases.json
+++ b/config/example_aliases.json
@@ -5,13 +5,19 @@
         "The keys must be valid commands, and can have simple, space separated arguments.   ",
         "Arguments passed by the user will be appended to any command arguments set here.   ",
         "                                                                                   ",
-        "For Example:                                                                       ",
+        "For example, (replace _cmd_name_ with the command of your choice):                 ",
+        "This will map _cmd_name_ to multiple aliases, alias1 & a_long_alias           ",
         {
-            "_command_name_":  ["alias1", "a_long_alias"],
-            "_cmd_ arg1 arg2 argN": ["short"],
-
-            "config set MusicBot": ["setcfg"]
-        }
+            "_cmd_name_":  ["alias1", "a_long_alias"]
+        },
+        "                                                                                   ",
+        "To alias a command with arguments, use something like the following:               ",
+        {
+            "_cmd_name_ arg1 arg2 argN": ["macro_alias"]
+        },
+        "Lets say you have the following alias: ",
+        {"config set MusicBot": ["setcfg"]},
+        "Users can use `setcfg DefaultVolume 0.40` to set default playback volume to 40%.   "
     ],
     
     "play": ["p"],
--- a/config/example_aliases.json
+++ b/config/example_aliases.json
@@ -1,4 +1,19 @@
 {
+    "--comment": [
+        "Aliases allow you to rename or shorten existing commands.                          ",
+        "Alias names may contain anything, except for spaces.                               ",
+        "The keys must be valid commands, and can have simple, space separated arguments.   ",
+        "Arguments passed by the user will be appended to any command arguments set here.   ",
+        "                                                                                   ",
+        "For Example:                                                                       ",
+        {
+            "_command_name_":  ["alias1", "a_long_alias"],
+            "_cmd_ arg1 arg2 argN": ["short"],
+
+            "config set MusicBot": ["setcfg"]
+        }
+    ],
+    
     "play": ["p"],
     "skip": ["s"],
     "queue": ["q"],
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -7035,6 +7035,59 @@ class MusicBot(discord.Client):
 
         return Response(codeblock.format(result))
 
+    @dev_only
+    async def cmd_makemarkdown(
+        self,
+        channel: MessageableChannel,
+        author: discord.Member,
+        cfg: str = "opts",
+    ) -> CommandResponse:
+        """
+        Command to generate markdown for options and permissions files.
+        Contents are generated from code and not pulled from the files!
+        """
+        valid_opts = ["opts", "perms"]
+        if cfg not in valid_opts:
+            opts = ", ".join([f"`{o}`" for o in valid_opts])
+            raise exceptions.CommandError("Option must be one of: %s" % (opts))
+
+        filename = "config_options.md"
+        msg_str = "Config options described in Markdown:\n"
+        if cfg == "perms":
+            filename = "config_permissions.md"
+            msg_str = "Permissions described in Markdown:\n"
+            config_md = self.permissions.register.export_markdown()
+        else:
+            config_md = self.config.register.export_markdown()
+
+        sent_to_channel = None
+
+        # TODO: refactor this in favor of safe_send_message doing it all.
+        with BytesIO() as fcontent:
+            fcontent.write(config_md.encode("utf8"))
+            fcontent.seek(0)
+            datafile = discord.File(fcontent, filename=filename)
+
+            try:
+                # try to DM. this could fail for users with strict privacy settings.
+                # or users who just can't get direct messages.
+                await author.send(msg_str, file=datafile)
+
+            except discord.errors.HTTPException as e:
+                if e.code == 50007:  # cannot send to this user.
+                    log.debug("DM failed, sending in channel instead.")
+                    sent_to_channel = await channel.send(
+                        msg_str,
+                        file=datafile,
+                    )
+                else:
+                    raise
+        if not sent_to_channel:
+            return Response(
+                "Sent a message with the requested config markdown.", delete_after=20
+            )
+        return None
+
     @owner_only
     async def cmd_checkupdates(self, channel: MessageableChannel) -> CommandResponse:
         """
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -247,6 +247,12 @@ class MusicBot(discord.Client):
         if self.config.enable_queue_history_global:
             await self.playlist_mgr.global_history.load()
 
+        # TODO: testing is needed to see if this would be required.
+        # See also:  https://github.com/aio-libs/aiohttp/discussions/6044
+        # aiohttp version must be at least 3.8.0 for the following to potentially work.
+        # Python 3.11+ might also be a requirement if CPython does not support start_tls.
+        # setattr(asyncio.sslproto._SSLProtocolTransport, "_start_tls_compatible", True)
+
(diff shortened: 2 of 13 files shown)


config/example_aliases.json
```diff
@@ -1,8 +1,23 @@
...
+    "--comment": [
+        "Aliases allow you to rename or shorten existing commands.                          ",
+        "Alias names may contain anything, except for spaces.                               ",
+        "The keys must be valid commands, and can have simple, space separated arguments.   ",
+        "Arguments passed by the user will be appended to any command arguments set here.   ",
+        "                                                                                   ",
+        "For Example:                                                                       ",
+        {
+            "_command_name_":  ["alias1", "a_long_alias"],
+            "_cmd_ arg1 arg2 argN": ["short"],
```
BabyBoySnow: This could use some clarification, maybe an example of what an argument is?

musicbot/bot.py
```diff
@@ -1564,13 +1573,20 @@ async def update_now_playing_status(self, set_offline: bool = False) -> None:
...
-        paused = sum(1 for p in self.players.values() if p.is_paused)
+        if self.config.status_include_paused:
+            paused = sum(1 for p in self.players.values() if p.is_paused)
+        else:
+            paused = 0
         total = len(self.players)
 
         def format_status_msg(player: Optional[MusicPlayer]) -> str:
+            if not self.config.status_include_paused:
+                p = sum(1 for p in self.players.values() if p.is_paused)
```
BabyBoySnow, line 1584: Shouldn't this be `if not p.is_paused`?

musicbot/bot.py
```diff
@@ -2631,6 +2647,9 @@ async def handle_player_inactivity(self, player: MusicPlayer) -> None:
         Manage a server-specific event timer when it's MusicPlayer becomes idle,
         if the bot is configured to do so.
         """
+        if self.logout_called:
+            return
```
BabyBoySnow, line 2651: Good call to add this here 👍

musicbot/bot.py
```diff
@@ -7016,24 +7062,85 @@ async def cmd_debug(
...
+        channel: MessageableChannel,
+        author: discord.Member,
+        cfg: str = "opts",
+    ) -> CommandResponse:
+        """
+        Command to generate markdown for options and permissions files.
+        Contents are generated from code and not pulled from the files!
+        """
+        valid_opts = ["opts", "perms"]
+        if cfg not in valid_opts:
```
BabyBoySnow, line 7104: Is it bad practice to make this all one line? `if cfg not "opts" or "perms":`

musicbot/config.py
```diff
@@ -918,6 +961,14 @@ def run_checks(self) -> None:
         if self.enable_local_media and not self.media_file_dir.is_dir():
             self.media_file_dir.mkdir(exist_ok=True)
 
+        if self.cookies_path.is_file():
+            log.warning(
+                "Cookies TXT file detected. MusicBot will pass them to yt-dlp.\n"
```
BabyBoySnow, line 980: I'd like a warning here about using personal daily accounts

musicbot/constants.py
```diff
@@ -88,6 +111,13 @@
 # Maximum number of seconds to wait for HEAD request on media files.
 DEFAULT_MAX_INFO_REQUEST_TIMEOUT: int = 10
 
+# MusicBot can pre-download the track up-next, while the current plays.
+# Typically not an issue, but makes more requests over all to the same media.
+# If it is an issue for some, this constant can disable the pre-downloads.
+PRE_DOWNLOAD_NEXT_TRACK: bool = True
```
BabyBoySnow: This should probably be an option as opposed to having to edit code

musicbot/downloader.py
```diff
@@ -85,13 +85,29 @@ def __init__(self, bot: "MusicBot") -> None:
         )
 
         # force ytdlp and HEAD requests to use the same UA string.
-        self.http_req_headers = {
-            "User-Agent": youtube_dl.utils.networking.random_user_agent()
-        }
+        # If the constant is set, use that, otherwise use dynamic selection.
+        if bot.config.ytdlp_user_agent:
+            ua = bot.config.ytdlp_user_agent
+            log.warning("Forcing YTDLP to use User Agent:  %s", ua)
```
BabyBoySnow, line 91: Talk with Justin about updating the docs and providing examples of valid UA's

update.py
```diff
@@ -343,20 +354,29 @@ def main() -> None:
...
+    # -suno is --short with --untracked-files=no
     status_unclean = subprocess.check_output(
-        [git_bin, "status", "--porcelain"], universal_newlines=True
+        [git_bin, "status", "-suno", "--porcelain"], universal_newlines=True
     )
-    if status_unclean:
+    if status_unclean.strip():
+        # TODO: Maybe offering a stash option here would not be so bad...
+        print(
+            "Detected the following files have been modified:\n"
```
BabyBoySnow, line 365: Maybe we should encode this like the bad install error message? I'll leave that choice up to you.

config/example_aliases.json
itsTheFae: Yeah, probably right.  Only real issue is trying to write comments in JSON, which doesn't have comments naturally.   
I'm open to suggestions on how to improve it.

musicbot/bot.py
itsTheFae: Not here.  This handles the case where counting paused players for dynamic status is disabled, but a custom status might still want to use the `{n_paused}` variable.   
So this independently counts the number of paused players so it can be displayed but wont impact the logic.

musicbot/bot.py
itsTheFae: Not bad practice afaik.  The list here also gets used in generating the error message to show what the valid opts are.

musicbot/constants.py
itsTheFae: I'd agree, but wasn't sure if I should make it an option since it was default behavior to pre-download before.  Though since it is also extra requests, I thought "maybe it ought to be easier to disable for testing sake?"   Not really sure though.

musicbot/bot.py
BabyBoySnow: Oh sorry I missed that

musicbot/constants.py
BabyBoySnow: That default behavior was added by Bork, originally it never did way back when lol. Making it optionally is good, less request is muy Bueno.

musicbot/downloader.py
itsTheFae: Technically speaking, any text, including nothing at all, is a valid UA.  They're entirely arbitrary.   If by valid, you mean matches an existing web-brower, we might be better off linking to Mozilla developer or some site that lists UAs?  

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent   

https://www.useragents.me/

musicbot/bot.py
BabyBoySnow: Oh okay, I simply saw the above line `if not self.config` and thought wait this isn't right lol, thought maybe you copy pasted it by accident.

config/example_aliases.json
BabyBoySnow: Hmm that does complicate things, I didn't know json couldn't contain comments, any chance of making it an ini file 😆

musicbot/downloader.py
BabyBoySnow: Oh that's probably fine then, I was just worried that doing something like 1234urmom might break things lol

config/example_aliases.json
itsTheFae: Eh, they probably could but I don't think I'll get around to redoing the format and parsing for this PR. lol  
Will admit that this basically yet another bolt-on and not really a rework.

musicbot/config.py
itsTheFae: I've updated the warning text here and added a sleep to make sure folks see it.

musicbot/downloader.py
itsTheFae: It typically wont break the bot.  But arbitrary as they may be, some servers may also tend to filter traffic based on headers being more close to expected norms.   So random strings could work for some servers and get you soft-blocked by automatic rules on others.

update.py
itsTheFae: Not sure we use the encoded message any more either. lol  
I figure its harmless enough.  If they want to ignore what files will be reset, they'll have no right to cry if they wipe out their own changes.

musicbot/downloader.py
BabyBoySnow: Adding the help links will most assuredly bring down complaints about something not working right 👍 I imagine most people will just copy paste an example

update.py
BabyBoySnow: Oops I think Jayden wanted that to stay there to deter people from downloading a .zip by just removing the check via Ctrl f'ing the error

itsTheFae: I don't know if I was the one who removed it.  I do remember removing the encoded keys for now long defunct email auth in the config.   
With how the bot pulls its version and checks for the git repo though, I figure folks would have a hard time getting around not being in a git repo these days.

itsTheFae: Though if it needs to be more clear, not to use the repo archive downloads then we can certainly do that too.

BabyBoySnow: It's probably fine I hope lol

PR #2417, opened 2024-07-29
Cookie management command & bug fix

itsTheFae:
- [ ] I have tested my changes on Python 3.8 or higher (only tested 3.10, but 3.8 should still work)

----

### Description

- Adds an owner-only command `setcookies` to allow managing cookies remotely.   
- Continue to download songs when the mostly optional header check request hits a timeout.
- Changes downloader to randomize UA strings for each extraction when UA is not static. (Fixes possible regression that may be causing more frequent sign-in errors.)

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -7354,6 +7354,12 @@ class MusicBot(discord.Client):
           When updating cookies, you must upload a file named cookies.txt
           If cookies are disabled, uploading will enable the feature.
           Uploads will delete existing cookies, including disabled cookies file.
+
+        WARNING:
+          Copying cookies can risk exposing your personal information or accounts,
+          and may result in account bans or theft if you are not careful.
+          It is not recommended due to these risks, and you should not use this
+          feature if you do not understand how to avoid the risks.
         """
         opt = opt.lower()
         if opt == "on":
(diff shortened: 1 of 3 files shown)


musicbot/bot.py
```diff
@@ -7338,6 +7338,101 @@ async def cmd_botversion(self) -> CommandResponse:
...
+            if self.downloader.cookies_enabled:
+                self.downloader.disable_ytdl_cookies()
+
+            if self.config.cookies_path.is_file():
+                try:
+                    self.config.cookies_path.rename(self.config.disabled_cookies_path)
+                except OSError as e:
+                    raise exceptions.CommandError(
+                        f"Could not rename cookies file due to error:  {str(e)}\n"
+                        "Cookies temporarilly disabled and will be re-enabled on next restart."
```
BabyBoySnow: temporarily

musicbot/downloader.py
```diff
@@ -168,6 +206,9 @@ async def get_url_headers(self, url: str) -> Dict[str, str]:
                         headers[new_key] = values
                     else:
                         headers[new_key] = values.pop()
+            except asyncio.exceptions.TimeoutError:
+                log.warning("Checking media headers failed due to timeout.")
+                headers = headers = {"X-HEAD-REQ-FAILED": "1"}
```
BabyBoySnow: Can't this just be `headers = {"X-HEAD-REQ-FAILED": "1"}` ?

PR #2432, opened 2024-10-18
feat: Update Docker and CI

pythoninthegrass:
- [x] I have tested my changes against the `dev` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [x] I have tested my changes on Python 3.8 or higher
- [ ] I have ensured my code is formatted using [Black](https://github.com/psf/black)

----

### Description

Original PR at #2431

> [!NOTE] 
> This PR includes recent updates from the `dev` branch. My specific contributions are focused on Docker and CI/CD improvements, as detailed in the commit "feat: update docker and CI/CD". 
>
> I did my best to rebase my feature branch against the upstream dev, but ended up coauthoring a handful of commits.
> 
> Hoping that since github says the branch can be merged automatically, that should be sufficient 🤞 
>

My key changes include:

- Updated `docker.yml` for multi-architecture builds and container registry pushes
- Refined `Dockerfile` with buildkit syntax and environment variable improvements
- Refactored `docker-compose.example.yml`
- Updated `docker-entrypoint.sh` to remove bashisms and control for word splitting
- Added `.dockerignore`, `.dive-ci`, and `.hadolint.yaml` for Docker build optimization and linting
- Updated `.gitignore` and added `.env.example`

These changes are aimed at improving the Docker-related aspects of the project and enhancing the CI/CD pipeline.

### Related issues (if applicable)

Closes #2412

PR #2438, opened 2024-11-18
Add github workflow for publishing a docker image

Gobidev:
This builds an image using the Dockerfile in the root dir of the repo and publishes it to GitHubs container registry.

~Let me know if you want me to add an example docker-compose.yml file to the readme or the docs on how to install the bot using the image.~ Nvm, I didn't realize there was a `docker-compose.example.yml` in the repo

The `Dockerfile` should also probably be updated to Python 3.10 at some point

### DIFF
--- a/.github/workflows/docker_image_release.yml
+++ b/.github/workflows/docker_image_release.yml
@@ -1,0 +1,41 @@
+name: Create and publish Docker image
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


BabyBoySnow:
There's some feedback in #2432 that I'd recommend checking out.

PR #2458, opened 2025-05-06
Add Concurrency to speed up downloads.

MuXodious:

- [] I have ensured my code is formatted using [Black](https://github.com/psf/black)

----

### Description
The PR merely adds the `concurrent_fragment_downloads` option configured at 16 concurrent downloads to the downloader.py to speed up downloads when someone from your server decides to have a 10 hour long sound playing on the background as a joke and the download reads "50 minutes" to be completed, ultimately defeating his tomfoolery.

https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/YoutubeDL.py#L511
https://github.com/yt-dlp/yt-dlp/blob/master/yt_dlp/downloader/fragment.py#L38

### DIFF
--- a/config/example_options.ini
+++ b/config/example_options.ini
@@ -301,14 +301,14 @@ YtdlpUserAgent =
 # To allow either IPv4 or v6, set this to:  *
 YtdlpSourceAddress = *
 
+#Set the number of concurrent fragment downloads.
+YtdlpConcurrentFrags = 1
+
 # Toggle the user block list feature, without emptying the block list.
 EnableUserBlocklist = yes
 
 # Enable the song block list feature, without emptying the block list.
 EnableSongBlocklist = no
-
-#Set the number of concurrent fragment downloads.
-ConcurrentDownloads = 1
 
 [Files]
 # An optional file path to a text file listing Discord User IDs, one per line.
--- a/config/example_options.ini
+++ b/config/example_options.ini
@@ -307,6 +307,9 @@ EnableUserBlocklist = yes
 # Enable the song block list feature, without emptying the block list.
 EnableSongBlocklist = no
 
+#Set the number of concurrent fragment downloads.
+ConcurrentDownloads = 1
+
 [Files]
 # An optional file path to a text file listing Discord User IDs, one per line.
 UserBlocklistFile = 
--- a/config/example_options.ini
+++ b/config/example_options.ini
@@ -301,7 +301,8 @@ YtdlpUserAgent =
 # To allow either IPv4 or v6, set this to:  *
 YtdlpSourceAddress = *
 
-#Set the number of concurrent fragment downloads.
+#Sets the number of threads to use for native hls and dash downloads. 
+#It is likely to increase network load exponentially, use caution.
 YtdlpConcurrentFrags = 1
 
 # Toggle the user block list feature, without emptying the block list.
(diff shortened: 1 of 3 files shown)


BabyBoySnow:
Concerned about overhead for lower end devices/servers, would recommended making this a configurable value, starting at 4 for older Pi devices, and power user can increase it as desired.

MuXodious:
> Concerned about overhead for lower end devices/servers, would recommended making this a configurable value, starting at 4 for older Pi devices, and power user can increase it as desired.

Sorry for the late reply.
Exactly. It should be in the configuration file with, preferably, "1" as the default value, but I'm not fluent in Python. Though, I can look into it, and perhaps learn smth along the way.

BabyBoySnow:
> > Concerned about overhead for lower end devices/servers, would recommended making this a configurable value, starting at 4 for older Pi devices, and power user can increase it as desired.
> 
> Sorry for the late reply. Exactly. It should be in the configuration file with, preferably, "1" as the default value, but I'm not fluent in Python. Though, I can look into it, and perhaps learn smth along the way.

All good, ironically, I just opened my computer and was starting to set this up, because I thought this was one of those made a github for this one thing, never open it again kinda situation lmao. 

Learning new things is always good, and I've heard that knowledge is power my whole life. If you need any help along the way, I'm more than happy to try my best to point you in the right direction. :D

MuXodious:
Well, I tried something. It appears to be working, based on a crude testing method (aka passing print functions everywhere).

itsTheFae:
Looks alright to me.  I would like to suggest renaming the config name to `YtdlpConcurrentFrags` to better associate the config with the yt-dlp settings.    Since MusicBot also has a thread pool it manages for download concurrency between servers the bot is in, the current name would be ambiguous.   

Thanks for contributing!   

P.S: nothing wrong with prints.  Feel free to abuse the log functions/levels for similar purposes. :)

MuXodious:
> Looks alright to me. I would like to suggest renaming the config name to `YtdlpConcurrentFrags` to better associate the config with the yt-dlp settings. Since MusicBot also has a thread pool it manages for download concurrency between servers the bot is in, the current name would be ambiguous.
> 
> Thanks for contributing!
> 
> P.S: nothing wrong with prints. Feel free to abuse the log functions/levels for similar purposes. :)

Aye, thank y'all for the feedback. I didn't know there was a thread pool meant for individual servers. Based on my understanding, and as a side note, changing yt-dlp's concurrent frags value from a default of 1 to anything else might exponentially increase the network load when multiple servers attempt to, lets say, pull a 10-hour video prank simultaneously.

Anyways, I changed the option name.

P.S: Indeed, I should also figure out logging and debugging functions/libraries, instead of doing this 😅
image

BabyBoySnow:
> changing yt-dlp's concurrent frags value from a default of 1 to anything else might exponentially increase the network load when multiple servers attempt to, lets say, pull a 10-hour video prank simultaneously.

That's good to note, maybe we can expand on the explanation comment in the example file? I do believe the default thread pool is set to 2, so hopefully that helps mitigate an issue like that. 

> P.S: Indeed, I should also figure out logging and debugging functions/libraries, instead of doing this 😅

Good ol' print statements have been a staple of development, and where previously all we had prior to adding the logging, nothing wrong with using it! Print to your delight! :)

PR #2474, opened 2026-03-30
Fix compatibility for Python 3.11+ by refactoring logger level creation

Kartus22:
Critical Compatibility Fix:

Without this change, the bot fails to initialize the logger on modern Python versions (specifically 3.11, 3.12, and the upcoming 3.13).

The previous exec() approach causes issues with how logging methods are injected into the logging.Logger class in newer Python environments.

By switching to a lambda with setattr, the bot now starts correctly on current stable Python releases.

Changes:

Replaced the string-based exec() call with a clean setattr using a lambda function.

Addressed the existing TODO comment in the code which suggested finding a better way to handle this.

Why this is better:

Stability: Ensures the bot is future-proof and runs on the latest Python versions provided by hosts like BisectHosting.

Security: Removes unsafe "un-pythonic" exec() calls.

Maintainability: Cleaner code that follows standard logging extension patterns.

- [] I have tested my changes against the `dev` branch (the latest developmental version), and this pull request is targeting that branch as a base
- [] I have ensured my code is formatted using [Black](https://github.com/psf/black)

----


therealjustsnow:
Please target the newer dev branch for PR's

Kartus22:
Done! I've updated the base branch to dev and resolved all merge conflicts as requested. The PRs should be ready for review now. :)

PR #2477, opened 2026-05-10
Add slash commands

therealjustsnow:

- [x] I have tested my changes on Python 3.9 or higher

----

### Description
Adds a new file for slash commands, some changes in bot.py required for this to work

### DIFF
--- a/musicbot/bot.py
+++ b/musicbot/bot.py
@@ -5279,10 +5279,11 @@ class MusicBot(commands.Bot):
         If no player is available the queue file will be removed if it exists.
         """
 
-        # Ensure player is not none and that it's not empty
-        if _player and len(_player.playlist) < 1:
+        # Ensure player is not none and that there is actually something queued.
+        # Note: current_entry is the playing track, not a queue entry — don't count it.
+        if _player and not _player.playlist.entries:
             raise exceptions.CommandError(
-                "There is nothing currently playing. Play something with a play command."
+                "There is nothing in the queue to clear."
             )
 
         # Try to gracefully clear the guild queue if we're not in a vc.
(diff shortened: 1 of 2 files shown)


chy1211:
Reviewed this because I wanted to run it on my own fork/deployment — overall really solid work, cleanly integrates with the existing permissions system and I didn't find any leftover debug cruft despite the commit history looking exploratory.

One bug though: `/restart` (all 5 subcommands) and `/shutdown` don't actually do anything. Both re-raise `RestartSignal`/`TerminateSignal` expecting them to reach `MusicBot.on_error` the same way the prefix-command path does, but exceptions raised inside `app_commands` callbacks go through discord.py's separate `CommandTree`/cog error path instead (wrapped in `CommandInvokeError`), which never reaches `Client.on_error`. So right now the bot shows "♻️ Restarting…" / "👋 Shutting down…" and then just... keeps running.

Fixed it by adding `cog_app_command_error` to the cog to catch it and replicate `on_error`'s handling (set `exit_signal`, call `logout()`). Verified with an isolated test exercising the exact error path discord.py would produce.

Opened #2483 with your commits cherry-picked as-is (authorship preserved) plus that fix on top, in case it's useful to the maintainers. Not trying to step on this — happy to close mine in favor of yours if you'd rather just pull the fix in here instead, just let me know.

therealjustsnow:
Hey thanks for catching the bug! I'll go ahead and close this, and leave yours open, easier than the other way around I think? 

PS: Sorry about the commit history 🤣
PPS: Is copilot comments on yours valid? I don't trust that guy 👀

chy1211:
All good on my end, appreciate you closing it in favor of mine — and honestly no apology needed for the commit history, the actual diff was clean, that's what matters 😄

Re: Copilot — fair skepticism in general, but I'll vouch for it this round. It left 4 comments on #2483 and I went and independently verified each one against the actual code before touching anything (didn't want to blindly trust a bot either). All 4 were real:

- `/help command:<dev-only-cmd>` bypassed the owner/dev check that the prefix version enforces — any user could pull up docs for the dev-only commands
- `/skip` vote counting silently never recorded the caller (slash calls `cmd_skip` with `message=None`, and the vote-add was gated behind `if message is not None`) — so vote-skip was effectively broken for anyone without InstaSkip
- `/queue page:` had no upper bound clamp, large page numbers just render an empty page
- `tree.sync()` at startup had no error handling, a transient Discord API hiccup could interrupt the rest of ready-up

Already pushed fixes for all of them onto #2483. So: not infallible, but this time it earned its keep.

